PR #1, opened 2019-05-03
Titan build script and other updates

yiwu-arbug:
Summary:
Adding a cmake script so we can build static lib for Titan:
* build static lib libtitan.a, titandb_bench, titandb_stress, and unit tests.
* the script should generate the same compile and link flags (by copying part of rocksdb/CMakeLists.txt into cmake/rocksdb_flags.cmake)
* See the new README.md file for build script usage

Other changes:
* updated README.md
* updated header paths
* remove titan_c.h and titan_c.cc, which is not being used anywhere
* add a .gitignore

Test Plan:
```
mkdir -p build; 
cd build; 
cmake .. -DROCKSDB_DIR=../rocksdb;
make -j
ctest -R titan
```


.gitignore
```diff
@@ -0,0 +1,5 @@
+*.a
+*.o
+*.swp
```
Connor1996: please add 
```
*_test
*_bench
*_stress

.idea/
.vscode/
```

README.md
```diff
@@ -1 +1,32 @@
...
\ No newline at end of file
+# Titan: A RocksDB Plugin to Reduce Write Amplification
+
+Titan is RocksDB Plugin for key-value separation, inspired by 
+[WiscKey](https://www.usenix.org/system/files/conference/fast16/fast16-papers-lu.pdf).
+For introduction and design details, see our
+[blog post](https://pingcap.com/blog/titan-storage-engine-design-and-implementation/).
+
+## Build and Test
+Titan rely on RocksDB source code to build. You need to checkout RocksDB source code locally,
```
Connor1996: can we use git submodule instead of local RocksDB?

Connor1996:
Also, we should set up CI and make sure not everyone can push commits into master directly

README.md
```diff
@@ -1 +1,32 @@
-# titan
\ No newline at end of file
+# Titan: A RocksDB Plugin to Reduce Write Amplification
+
+Titan is RocksDB Plugin for key-value separation, inspired by
```
Connor1996: Titan is RocksDB Plugin for key-value separation.
```suggestion
Titan is a RocksDB Plugin for key-value separation, inspired by 
```

README.md
Connor1996:
```suggestion
Titan relies on RocksDB source code to build. You need to checkout RocksDB source code locally,
```

.gitignore
yiwu-arbug: sure. btw it is recommended to build in a standalone directory (e.g `mkdir -p build; cd build; cmake ..; make`) so I ignore the build/ directory only.

README.md
yiwu-arbug: For use in rust-rocksdb repo, i'll make Titan a submodule. Since we already have the rocksdb submodule there, having another nested rocksdb inside titan submodule sounds duplicated. I have been debating with myself with some of the options:
1. require a local RocksDB copy.
2. copy rocksdb internal headers into include/ directory. The downside is we would need to maintain those headers to be consistent with rocksdb.
3. make rocksdb a submodule of Titan, and remove the rocksdb submodule from rust-rocksdb. The downside is whenever we update rocksdb there's one more step to do (i.e. update in pingcap/rocksdb repo->update in titan repo->and update in rust-rocksdb repo).

Which do you think will work better?

yiwu-arbug:
@Connor1996 I'll setup CI in another PR.

README.md
Connor1996: okay, let's keep this way

PR #2, opened 2019-05-07
Add travis config

yiwu-arbug:
Adding travis config to trigger CI build. Adjust cmake script accordingly.


yiwu-arbug:
ASAN and TSAN build failure is expected. will fix later.

PR #7, opened 2019-05-14
Add code coverage CI

yiwu-arbug:
use `lcov` to generate code coverage report together with GCC build, and send result to codecov. Also some minor refactor to travis config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,13 @@ addons:
     sources:
       - ubuntu-toolchain-r-test
     packages:
-      - g++-8
+      - g++-7
       - libgflags-dev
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -10,13 +10,13 @@ addons:
     sources:
       - ubuntu-toolchain-r-test
     packages:
-      - g++-8
+      - g++-7
```
yiwu-arbug: In travis the newest lcov is 1.12, and it is not compatible with gcc8. So I change to gcc7.

Connor1996:
wow, we have a coverage of 84.77% ~

PR #8, opened 2019-05-14
change codecov comment style

yiwu-arbug:
codecov comment on every PR, which is a bit annoying. try to disable it. 

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,1 @@
+comment: off
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,3 @@
 comment:
-  layout: "diff, files"
-  behavior: new
+  layout: "diff"
+  behavior: default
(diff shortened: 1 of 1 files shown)


Connor1996:
I think invoking it on every PR is fine🤣So how to get the report after disabling it?

yiwu-arbug:
click on the codecov badge icon and select pulls. I think we don't necessary need to worry about code coverage for every PR?
Screen Shot 2019-05-14 at 8 41 14 PM

Connor1996:
The codecov report enforces us to care about the coverage and make sure at least not letting it drop when merging PR, especially for some breaking changes. If we don't trigger it actively, I think we would hardly to click on that icon.
And TiDB and PD show the codecov report on every PR.

yiwu-arbug:
Changed the config to keep the codecov comment but only showing a summary.

PR #9, opened 2019-05-14
Add clang-format script and CI job

yiwu-arbug:
Adding scripts/format-diff.sh to use clang-format to format code changes since diverge from master. Also adding travis CI job to check if the code needs to format. Also format all existing code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,16 +35,16 @@ install:
   - if [ "${COMPILER}" == gcc7 ]; then
       CC=gcc-7;
       CXX=g++-7;
-      COVERAGE_OPT="-DCODE_COVERAGE=ON";
+      export COVERAGE_OPT="-DCODE_COVERAGE=ON";
     fi
   - if [ ! -z "${BUILD_TYPE}" ]; then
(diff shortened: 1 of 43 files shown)


.travis.yml
```diff
@@ -26,28 +28,31 @@ matrix:
...
-      TOOLS_OPT="-DWITH_TITAN_TOOLS=OFF";
+      export SANITIZER_OPT="-DWITH_${SANITIZER}=ON";
+      export TOOLS_OPT="-DWITH_TITAN_TOOLS=OFF";
     fi
 
 script:
-  - cmake . -L -DROCKSDB_DIR=./rocksdb -DTRAVIS=ON ${BUILD_OPT} ${SANITIZER_OPT} ${TOOLS_OPT} ${COVERAGE_OPT}
-  - make -j4
-  - ctest -R titan
+  - if [ -z "${FORMATTER}" ]; then
```
Connor1996: should we make format check before all tests? Once the format check is failed, running other tests is a waste of CI compute source.

yiwu-arbug: All the jobs will run in parallel until finish anyway. and IMO having tests pass is more important than formatter?

PR #10, opened 2019-05-15
fix test config for min-blob-size

tabokie:
Add an uint64 test flag `min-blob-size` for Titan in titandb_bench and titandb_stress tests.

### DIFF
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -735,11 +735,14 @@ DEFINE_uint64(blob_db_bytes_per_sync, 0, "Bytes to sync blob file at.");
 DEFINE_uint64(blob_db_file_size, 256 * 1024 * 1024,
               "Target size of each blob file.");
 
+DEFINE_uint64(blob_db_min_blob_size, 0,
+              "Smallest blob to store in a file. Blob smaller than this "
+              "will be inlined with the key in the LSM tree.");
+
 // Titan Options
 DEFINE_bool(use_titan, true, "Open a Titan instance.");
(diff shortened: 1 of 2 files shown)


tabokie:
PTAL @Connor1996

tools/db_bench_tool.cc
```diff
@@ -730,15 +730,19 @@ DEFINE_uint64(blob_db_max_ttl_range, 86400,
 DEFINE_uint64(blob_db_ttl_range_secs, 3600,
               "TTL bucket size to use when creating blob files.");
 
-DEFINE_uint64(blob_db_min_blob_size, 0,
```
Connor1996: you should not delete this, it is for blobDB

tools/db_bench_tool.cc
```diff
@@ -730,15 +730,19 @@ DEFINE_uint64(blob_db_max_ttl_range, 86400,
...
 DEFINE_uint64(blob_db_bytes_per_sync, 0, "Bytes to sync blob file at.");
 
 DEFINE_uint64(blob_db_file_size, 256 * 1024 * 1024,
               "Target size of each blob file.");
 
+// Titan Options
+DEFINE_bool(use_titan, true, "Open a Titan instance.");
+
+// Shared Options
+DEFINE_uint64(min_blob_size, 0,
```
Connor1996:
```suggestion
DEFINE_uint64(titan_db_min_blob_size, 0,
```

tabokie:
PTAL @Connor1996

PR #11, opened 2019-05-22
Add blob-run-mode for downgrading Titan

tabokie:
blob-run-mode is used for smoothly downgrading Titan to native RocksDB.
There are three modes for blob-run-mode option:
* "normal": Titan process read/write as normal.
* "read-only": Titan stop writing value into blob log during flush and compaction. Existing values in blob log is still readable.
* "fallback": On flush and compaction, Titan will convert blob index into real value, by reading from blob log, and store the value in SST file.

Note that currently TitanCFOption is cached everywhere, thus this PR uses a lousy value broadcast method. Future work should make mutable part of TitanCFOption easier to deal with.
Also, add internal mutex for VersionSet and fix some missing option setter/getter for TitanDB.

### DIFF
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -32,11 +32,11 @@ struct TitanDBOptions : public DBOptions {
 };
 
 enum class TitanBlobRunMode {
-  kNormal = 0,     // Titan process read/write as normal
-  kReadOnly = 1,  // Titan stop writing value into blob log during flush 
-                  // and compaction. Existing values in blob log is still 
+  kNormal = 0,    // Titan process read/write as normal
+  kReadOnly = 1,  // Titan stop writing value into blob log during flush
+                  // and compaction. Existing values in blob log is still
                   // readable and garbage collected.
-  kFallback = 2,  // On flush and compaction, Titan will convert blob 
+  kFallback = 2,  // On flush and compaction, Titan will convert blob
                   // index into real value, by reading from blob log, 
                   // and store the value in SST file.
 };
@@ -88,7 +88,7 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // Default: 8MB
   uint64_t merge_small_file_threshold{8 << 20};
 
-  // The blob running mode used to turn off Titan. This option is only 
+  // The blob running mode used to turn off Titan. This option is only
   // valid for default column family using key-value seperation.
   //
   // Default: kNormal
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -2,6 +2,8 @@
 
 #include "rocksdb/options.h"
 
+#include <unordered_map>
+
 namespace rocksdb {
 namespace titandb {
 
@@ -40,6 +42,14 @@ enum class TitanBlobRunMode {
                   // index into real value, by reading from blob log,
                   // and store the value in SST file.
 };
+
+struct TitanOptionsHelper {
+  static std::unordered_map<std::string, TitanBlobRunMode>
+      blob_run_mode_string_map;
+};
+
+static auto& blob_run_mode_string_map =
+    TitanOptionsHelper::blob_run_mode_string_map;
 
 struct TitanCFOptions : public ColumnFamilyOptions {
   // The smallest value to store in blob files. Value smaller than
@@ -88,8 +98,7 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // Default: 8MB
   uint64_t merge_small_file_threshold{8 << 20};
 
-  // The blob running mode used to turn off Titan. This option is only
-  // valid for default column family using key-value seperation.
+  // The mode used to process blob file.
   //
   // Default: kNormal
   TitanBlobRunMode blob_run_mode{TitanBlobRunMode::kNormal};
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -37,7 +37,7 @@ enum class TitanBlobRunMode {
                   // and compaction. Existing values in blob log is still
                   // readable and garbage collected.
   kFallback = 2,  // On flush and compaction, Titan will convert blob
-                  // index into real value, by reading from blob log, 
+                  // index into real value, by reading from blob log,
                   // and store the value in SST file.
 };
 
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -51,6 +51,9 @@ struct TitanOptionsHelper {
 static auto& blob_run_mode_string_map =
     TitanOptionsHelper::blob_run_mode_string_map;
 
+struct ImmutableTitanCFOptions;
+struct MutableTitanCFOptions;
+
 struct TitanCFOptions : public ColumnFamilyOptions {
   // The smallest value to store in blob files. Value smaller than
   // this threshold will be inlined in base DB.
@@ -106,6 +109,8 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   TitanCFOptions() = default;
   explicit TitanCFOptions(const ColumnFamilyOptions& options)
       : ColumnFamilyOptions(options) {}
+  explicit TitanCFOptions(const ImmutableTitanCFOptions&,
+                          const MutableTitanCFOptions&);
 
   TitanCFOptions& operator=(const ColumnFamilyOptions& options) {
     *dynamic_cast<ColumnFamilyOptions*>(this) = options;
@@ -113,6 +118,38 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   }
 
   std::string ToString() const;
+};
+
+struct ImmutableTitanCFOptions {
+  uint64_t min_blob_size{4096};
+  CompressionType blob_file_compression{kNoCompression};
+  uint64_t blob_file_target_size{256 << 20};
+  std::shared_ptr<Cache> blob_cache;
+  uint64_t max_gc_batch_size{1 << 30};
+  uint64_t min_gc_batch_size{512 << 20};
+  float blob_file_discardable_ratio{0.5};
+  float sample_file_size_ratio{0.1};
+  uint64_t merge_small_file_threshold{8 << 20};
+
+  ImmutableTitanCFOptions() = default;
+  explicit ImmutableTitanCFOptions(const TitanCFOptions& opts)
+      : min_blob_size(opts.min_blob_size),
+        blob_file_compression(opts.blob_file_compression),
+        blob_file_target_size(opts.blob_file_target_size),
+        blob_cache(opts.blob_cache),
+        max_gc_batch_size(opts.max_gc_batch_size),
+        min_gc_batch_size(opts.min_gc_batch_size),
+        blob_file_discardable_ratio(opts.blob_file_discardable_ratio),
+        sample_file_size_ratio(opts.sample_file_size_ratio),
+        merge_small_file_threshold(opts.merge_small_file_threshold) {}
+};
+
+struct MutableTitanCFOptions {
+  TitanBlobRunMode blob_run_mode{TitanBlobRunMode::kNormal};
+
+  MutableTitanCFOptions() = default;
+  explicit MutableTitanCFOptions(const TitanCFOptions& opts)
+      : blob_run_mode(opts.blob_run_mode) {}
 };
 
 struct TitanOptions : public TitanDBOptions, public TitanCFOptions {
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -31,6 +31,16 @@ struct TitanDBOptions : public DBOptions {
   }
 };
 
+enum class TitanBlobRunMode {
+  kNormal = 0,     // Titan process read/write as normal
+  kReadOnly = 1,  // Titan stop writing value into blob log during flush 
+                  // and compaction. Existing values in blob log is still 
+                  // readable and garbage collected.
+  kFallback = 2,  // On flush and compaction, Titan will convert blob 
+                  // index into real value, by reading from blob log, 
+                  // and store the value in SST file.
+};
+
 struct TitanCFOptions : public ColumnFamilyOptions {
   // The smallest value to store in blob files. Value smaller than
   // this threshold will be inlined in base DB.
@@ -53,30 +63,36 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // Default: nullptr
   std::shared_ptr<Cache> blob_cache;
 
-  // Max batch size for gc
+  // Max batch size for GC.
   //
   // Default: 1GB
   uint64_t max_gc_batch_size{1 << 30};
 
-  // Min batch size for gc
+  // Min batch size for GC.
   //
   // Default: 512MB
   uint64_t min_gc_batch_size{512 << 20};
 
-  // The ratio of how much discardable size of a blob file can be GC
+  // The ratio of how much discardable size of a blob file can be GC.
   //
   // Default: 0.5
   float blob_file_discardable_ratio{0.5};
 
-  // The ratio of how much size of a blob file need to be sample before GC
+  // The ratio of how much size of a blob file need to be sample before GC.
   //
   // Default: 0.1
   float sample_file_size_ratio{0.1};
 
-  // The blob file size less than this option will be mark gc
+  // The blob file size less than this option will be mark GC.
   //
   // Default: 8MB
   uint64_t merge_small_file_threshold{8 << 20};
+
+  // The blob running mode used to turn off Titan. This option is only 
+  // valid for default column family using key-value seperation.
+  //
+  // Default: kNormal
+  TitanBlobRunMode blob_run_mode{TitanBlobRunMode::kNormal};
 
   TitanCFOptions() = default;
   explicit TitanCFOptions(const ColumnFamilyOptions& options)
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -121,17 +121,8 @@ struct TitanCFOptions : public ColumnFamilyOptions {
 };
 
 struct ImmutableTitanCFOptions {
-  uint64_t min_blob_size{4096};
-  CompressionType blob_file_compression{kNoCompression};
-  uint64_t blob_file_target_size{256 << 20};
-  std::shared_ptr<Cache> blob_cache;
-  uint64_t max_gc_batch_size{1 << 30};
-  uint64_t min_gc_batch_size{512 << 20};
-  float blob_file_discardable_ratio{0.5};
-  float sample_file_size_ratio{0.1};
-  uint64_t merge_small_file_threshold{8 << 20};
+  ImmutableTitanCFOptions() : ImmutableTitanCFOptions(TitanCFOptions()) {}
 
-  ImmutableTitanCFOptions() = default;
   explicit ImmutableTitanCFOptions(const TitanCFOptions& opts)
       : min_blob_size(opts.min_blob_size),
         blob_file_compression(opts.blob_file_compression),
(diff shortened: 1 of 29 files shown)


tabokie:
PTAL, @yiwu-arbug

src/version_set.h
```diff
@@ -56,6 +56,7 @@ class VersionSet {
 
   // REQUIRES: mutex is held
```
Connor1996: remove these comments

include/titan/options.h
```diff
@@ -53,31 +63,37 @@ struct TitanCFOptions : public ColumnFamilyOptions {
...
   float sample_file_size_ratio{0.1};
 
-  // The blob file size less than this option will be mark gc
+  // The blob file size less than this option will be mark GC.
   //
   // Default: 8MB
   uint64_t merge_small_file_threshold{8 << 20};
 
+  // The blob running mode used to turn off Titan. This option is only
+  // valid for default column family using key-value seperation.
```
Connor1996: why is only valid for the default column?

src/table_builder.h
```diff
@@ -49,6 +52,7 @@ class TitanTableBuilder : public TableBuilder {
   std::unique_ptr<BlobFileHandle> blob_handle_;
   std::shared_ptr<BlobFileManager> blob_manager_;
   std::unique_ptr<BlobFileBuilder> blob_builder_;
+  std::weak_ptr<VersionSet> vset_ref_;
```
yiwu-arbug: what's the rationale to use weak_ptr?

src/table_builder.cc
```diff
@@ -12,7 +12,31 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
     return;
   }
 
-  if (ikey.type != kTypeValue || value.size() < cf_options_.min_blob_size) {
+  std::shared_ptr<VersionSet> vset;
+  if (ikey.type != kTypeValue &&
+      cf_options_.blob_run_mode == TitanBlobRunMode::kFallback &&
+      (vset = vset_ref_.lock()) != nullptr) {
```
yiwu-arbug: converting weak_ptr to shared_ptr for every key doesn't look efficient to me, though I don't know how much the penalty is. Since `vset` always outlive table factory (since `TitanDBImpl::Close()` close rocksdb before destruct itself), I think we can have hold raw pointer in `TitanTableBuilder`.

src/table_builder.cc
```diff
@@ -12,7 +12,31 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+    Slice copy = value;
+    uint64_t cf_id = GetTableProperties().column_family_id;
```
yiwu-arbug: getting CF id from table property looks hacky to me. Can we pass CF id to table factory on `TitanDBImpl::Open()` then pass the id to table builder from table factory?

src/table_builder.cc
```diff
@@ -12,7 +12,31 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+    PinnableSlice buffer;
+
+    auto storage = vset->GetBlobStorage(cf_id).lock();
+
+    ReadOptions options;  // dummy option
+    s = storage->Get(options, index, &record, &buffer);
+    if (s.ok()) {
+      base_builder_->Add(key, record.value);
+      return;
+    }
```
yiwu-arbug: set `status_ = s` on error?

src/table_builder.cc
```diff
@@ -12,7 +12,31 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+    BlobRecord record;
+    PinnableSlice buffer;
+
+    auto storage = vset->GetBlobStorage(cf_id).lock();
+
+    ReadOptions options;  // dummy option
+    s = storage->Get(options, index, &record, &buffer);
+    if (s.ok()) {
+      base_builder_->Add(key, record.value);
+      return;
```
yiwu-arbug: early return is not very readable. Use if-else instead?
```
if (type == kTypeBlobIndex && blob_run_mode == kFallback) {
   // get blob and convert to normal value
} else if (type == kTypeValue && blob_run_mode == kNormal && value.size() >= min_blob_size) {
   // write blob file and add blob index
} else {
  basd_builder_->Add(key, value)
}
```

src/table_builder.cc
```diff
@@ -12,7 +12,31 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+
+    BlobRecord record;
+    PinnableSlice buffer;
+
+    auto storage = vset->GetBlobStorage(cf_id).lock();
+
+    ReadOptions options;  // dummy option
+    s = storage->Get(options, index, &record, &buffer);
+    if (s.ok()) {
+      base_builder_->Add(key, record.value);
```
yiwu-arbug: you need to set the key type to `kTypeValue` here. (look at lines around `AppendInternalKey` below for example.

src/table_builder.h
tabokie: Originally I was trying to make it comfortable to pass a null pointer as vset in table-builder test. And also try not to make temporary table-builder instance hold strong reference.

src/table_builder.cc
tabokie: Okay

include/titan/options.h
tabokie: Non-Default CF do not adopt key-value separated approach for their value length are small. It's a upper level logic though, I shouldn't mention it here

tabokie:
@yiwu-arbug The muted fix is necessary for this PR, since it's irrational to pass a muted reference to table builder. So to take that out this PR has to be postponed.

include/titan/options.h
Connor1996: This constraint is only for TiKV. For other product, it may use other column families.

Connor1996:
@tabokie You just need to extract that logic to a new PR and keep this PR unchanged(no need to remove the logic). Once the new PR is merged to master, merge master for this PR and the diff would be clean.

src/table_builder.cc
tabokie: Builder holds an id data, I missed it.

src/table_builder.cc
tabokie: Here I was actually making sure that index value is not lost if blob storage is corrupted for now. But I guess that's a pointless act too.

src/table_factory.h
```diff
@@ -45,6 +48,8 @@ class TitanTableFactory : public TableFactory {
 
   void* GetOptions() override { return base_factory_->GetOptions(); }
 
+  TitanCFOptions* GetCFOptions() { return &cf_options_; }
```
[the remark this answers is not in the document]
tabokie: Should we make the change reflected in cf_options_ whenever used? Since passing a bool is not scalable, and the `GetPrintableTableOptions` will return stale data too.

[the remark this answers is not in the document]
yiwu-arbug: Let's add two new structs, `ImmutableTitanCFOptions` and `MutableTitanOptions`, which contain non-changeable/changeable options from `TitanCFOptions`.

src/titan_db_test.cc
```diff
@@ -606,6 +607,40 @@ TEST_F(TitanDBTest, BlobFileCorruptionErrorHandling) {
 }
 #endif  // !NDEBUG
 
+TEST_F(TitanDBTest, BlobRunModeBasic) {
+  Open();
```
yiwu-arbug: explicitly disable GC for the test?

src/titan_db_test.cc
```diff
@@ -606,6 +607,40 @@ TEST_F(TitanDBTest, BlobFileCorruptionErrorHandling) {
...
+  std::unordered_map<std::string, std::string> opts;
+  opts["blob-run-mode"] = "read-only";
+  db_->SetOptions(opts);
+  for (uint64_t i = kNumEntries + 1; i <= kNumEntries * 2; i++) {
+    Put(i, &data);
+  }
+  ASSERT_EQ(kNumEntries * 2, data.size());
+  VerifyDB(data);
+  Flush();
+  VerifyDB(data);
```
yiwu-arbug: You should verify:
1. The number of blob files is not growing
2. value type for key 100-199 are kTypeValue (you can use this helper function: https://github.com/facebook/rocksdb/blob/master/include/rocksdb/utilities/debug.h#L39)

src/titan_db_test.cc
```diff
@@ -606,6 +607,40 @@ TEST_F(TitanDBTest, BlobFileCorruptionErrorHandling) {
...
+
+  opts["blob-run-mode"] = "fallback";
+  db_->SetOptions(opts);
+  for (uint64_t i = kNumEntries + 1; i <= kNumEntries * 2; i++) {
+    Put(i, &data);
+  }
+  ASSERT_EQ(kNumEntries * 2, data.size());
+  VerifyDB(data);
+  Flush();
+  VerifyDB(data);
```
yiwu-arbug: ditto

tabokie:
The mutability fix is contained to a minimun scope, PTAL @yiwu-arbug

src/table_builder_test.cc
```diff
@@ -80,8 +80,8 @@ class TableBuilderTest : public testing::Test {
     db_options_.dirname = tmpdir_;
     cf_options_.min_blob_size = kMinBlobSize;
     blob_manager_.reset(new FileManager(db_options_));
-    table_factory_.reset(
-        new TitanTableFactory(db_options_, cf_options_, blob_manager_));
+    table_factory_.reset(new TitanTableFactory(db_options_, cf_options_,
+                                               blob_manager_, nullptr));
```
yiwu-arbug: pass a fake `VersionSet` here?

src/titan_db_test.cc
```diff
@@ -606,6 +608,70 @@ TEST_F(TitanDBTest, BlobFileCorruptionErrorHandling) {
...
+    Put(i, &data);
+  }
+  begin_key = GenKey(1);
+  end_key = GenKey(kNumEntries);
+  ASSERT_EQ(kNumEntries, data.size());
+  VerifyDB(data);
+  Flush();
+  auto blob = GetBlobStorage();
+  num_blob_files = blob.lock()->NumBlobFiles();
+  VerifyDB(data);
```
yiwu-arbug: Also verify part of the keys has type `kTypeBlobIndex`?

src/table_builder.cc
```diff
@@ -12,19 +12,46 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
     return;
   }
 
-  if (ikey.type != kTypeValue || value.size() < cf_options_.min_blob_size) {
+  if (ikey.type == kTypeBlobIndex &&
+      cf_options_.blob_run_mode == TitanBlobRunMode::kFallback &&
+      vset_ != nullptr) {
```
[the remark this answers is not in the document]
tabokie: What about table_builder_test which requires a vset pointer passed in? There mocking a VersionSet could be troublesome.

src/db_impl.cc
```diff
@@ -496,6 +494,25 @@ Options TitanDBImpl::GetOptions(ColumnFamilyHandle* column_family) const {
   return options;
 }
 
+Status TitanDBImpl::SetOptions(
```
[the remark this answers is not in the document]
tabokie: We can't do that easily. TitanOptions does not derived from Options.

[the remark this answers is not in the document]
yiwu-arbug: ok, I didn't notice it.

src/db_impl.cc
```diff
@@ -496,6 +494,31 @@ Options TitanDBImpl::GetOptions(ColumnFamilyHandle* column_family) const {
...
+    ColumnFamilyHandle* column_family,
+    const std::unordered_map<std::string, std::string>& new_options) {
+  auto opts = new_options;
```
yiwu-arbug: we should hold `mutex_` lock here I think, for accessing `titan_table_factory_`.

src/db_impl.cc
```diff
@@ -496,6 +494,31 @@ Options TitanDBImpl::GetOptions(ColumnFamilyHandle* column_family) const {
...
+  if (p != opts.end()) {
+    TitanBlobRunMode mode = TitanBlobRunMode::kNormal;
+    auto pm = blob_run_mode_string_map.find(p->second);
+    if (pm == blob_run_mode_string_map.end()) {
+      return Status::InvalidArgument("No blob-run-mode defined for " +
+                                     p->second);
+    } else {
+      mode = pm->second;
+    }
+    auto table_factory = titan_table_factory_[column_family->GetID()];
```
yiwu-arbug: `auto&` is better? to avoid one shared_ptr construction.

src/db_impl.h
```diff
@@ -149,7 +155,9 @@ class TitanDBImpl : public TitanDB {
   Statistics* stats_;
 
   std::unordered_map<uint32_t, std::shared_ptr<TableFactory>>
-      original_table_factory_;
+      base_table_factory_;
+  std::unordered_map<uint32_t, std::shared_ptr<TitanTableFactory>>
+      titan_table_factory_;
```
yiwu-arbug: Do we update these two map after dropping a CF? I don't see related logic.

src/options.cc
```diff
@@ -11,6 +11,19 @@
...
+      blob_file_compression(immutable_opts.blob_file_compression),
+      blob_file_target_size(immutable_opts.blob_file_target_size),
+      blob_cache(immutable_opts.blob_cache),
+      max_gc_batch_size(immutable_opts.max_gc_batch_size),
+      min_gc_batch_size(immutable_opts.min_gc_batch_size),
+      blob_file_discardable_ratio(immutable_opts.blob_file_discardable_ratio),
+      sample_file_size_ratio(immutable_opts.sample_file_size_ratio),
+      merge_small_file_threshold(immutable_opts.merge_small_file_threshold),
+      blob_run_mode(mutable_opts.blob_run_mode) {}
+
```
yiwu-arbug: update `ToString()` to print `blob_run_mode`?

src/table_factory.h
```diff
@@ -45,15 +49,23 @@ class TitanTableFactory : public TableFactory {
 
   void* GetOptions() override { return base_factory_->GetOptions(); }
 
+  void SetBlobRunMode(TitanBlobRunMode mode) {
```
yiwu-arbug: Either add comment saying holding `TitanDBImpl::mutex_` lock is required, or add a mutex to guard `mutable_cf_options_`.

tabokie:
> Almost there! Some more comments.
> 
> BTW does this PR really depend on #13?

It is. We do not want to pass a mutex reference to TableBuilder since it is accessing vset.

src/db_impl.h
tabokie: adding them now.

src/table_factory.h
```diff
@@ -45,15 +49,24 @@ class TitanTableFactory : public TableFactory {
 
   void* GetOptions() override { return base_factory_->GetOptions(); }
 
+  // REQUIRE: mutex held
+  void SetBlobRunMode(TitanBlobRunMode mode) {
+    MutableTitanCFOptions opts(mutable_cf_options_);
+    opts.blob_run_mode = mode;
+    mutable_cf_options_ = opts;
```
yiwu-arbug: simply `mutable_cf_options_.blob_run_mode = mode`?

src/table_factory.h
```diff
@@ -45,15 +49,24 @@ class TitanTableFactory : public TableFactory {
 
   void* GetOptions() override { return base_factory_->GetOptions(); }
 
+  // REQUIRE: mutex held
```
yiwu-arbug: Sorry. After a second though I think adding a mutex to guard `mutable_cf_options_` is a better idea. rocksdb flush/compaction job can run in concurrent and call `NewTableBuilder`. Both `NewTableBuilder` and `SetBlobRunMode` need to take the mutex.

src/db_impl.cc
```diff
@@ -496,6 +507,32 @@ Options TitanDBImpl::GetOptions(ColumnFamilyHandle* column_family) const {
...
+    } else {
+      mode = pm->second;
+    }
+    MutexLock l(&mutex_);
+    auto& table_factory = titan_table_factory_[column_family->GetID()];
+    table_factory->SetBlobRunMode(mode);
+    opts.erase(p);
+  }
+  if (opts.size() > 0) {
+    return db_->SetOptions(column_family, opts);
```
yiwu-arbug: Let's call `db_->SetOptions()` before handling `blob_run_mode`. In case `db_->SetOptions()` fail we should make the call fail as a whole.

src/options.cc
```diff
@@ -36,9 +36,18 @@ std::string TitanCFOptions::ToString() const {
...
+  snprintf(buf, sizeof(buf), "blob_run_mode = %s\n",
+           blob_run_mode_to_string[blob_run_mode].c_str());
+  res += buf;
   return res;
 }
 
+std::map<TitanBlobRunMode, std::string>
+    TitanOptionsHelper::blob_run_mode_to_string = {
+        {TitanBlobRunMode::kNormal, "normal"},
+        {TitanBlobRunMode::kReadOnly, "read_only"},
```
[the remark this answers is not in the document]
tabokie: should be named "read_only" like RocksDB, fix string_map now.

src/table_factory.h
tabokie: Isn't that minor compared to frequent use of mutex when accessing different container? Could we optimize that as well, e.g. use multiple spinlocks / make segmented map.

yiwu-arbug: what do you mean by "accessing different container"? Adding a mutex is not for performance, it is to make access of `blob_run_mode` thread-safe. RocksDB flush/compaction has no way to take TitanDBImpl::mutex_ before calling `NewTableBuilder`.

tabokie: got it.

src/db_impl.cc
```diff
@@ -510,27 +512,34 @@ Options TitanDBImpl::GetOptions(ColumnFamilyHandle* column_family) const {
...
+  TitanBlobRunMode mode = TitanBlobRunMode::kNormal;
+  auto pm = blob_run_mode_string_map.find(blob_run_mode_string);
+  if (pm == blob_run_mode_string_map.end()) {
+    return Status::InvalidArgument("No blob_run_mode defined for " +
+                                   blob_run_mode_string);
   } else {
-    return Status::OK();
+    mode = pm->second;
   }
+  mutex_.Lock();
```
yiwu-arbug: This is more readable:
```
{
  MutexLock l(&mutex_);
  auto& table_factory = titan_table_factory_[column_family->GetID()];
  table_factory->SetBlobRunMode(mode);
}
```

src/table_builder.cc
```diff
@@ -12,19 +12,47 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+      return;
+    }
 
-  std::string index_value;
-  AddBlob(ikey.user_key, value, &index_value);
-  if (!ok()) return;
+    BlobRecord record;
+    PinnableSlice buffer;
+
+    auto storage = vset_->GetBlobStorage(cf_id_).lock();
```
yiwu-arbug: instead of have table builder hold the version set, let's have table builder only holding blob storage of the CF. That way this patch don't depend on #13.

tabokie: Still need to pass mutex reference to table factory though.

src/table_factory.cc
```diff
@@ -20,12 +20,19 @@ TableBuilder* TitanTableFactory::NewTableBuilder(
     WritableFileWriter* file) const {
   std::unique_ptr<TableBuilder> base_builder(
       base_factory_->NewTableBuilder(options, column_family_id, file));
-  return new TitanTableBuilder(column_family_id, db_options_, cf_options_,
-                               std::move(base_builder), blob_manager_);
+  MutexLock l(&mutex_);
```
yiwu-arbug: something like this is better:
```
std::weak_ptr<BlobStorage> blob_storage;
{
  MutexLock db_l(&db_mutex_);
  blob_storage = vset_->GetBlobStorage(column_family_id);
}
TitanCFOptions cf_options;
{ 
  MutexLock l(&mutex_);
  cf_options = TitanCFOptions(immutable_cf_options_, mutable_cf_options_);
}
return new ...;

yiwu-arbug:
Mind run titandb_bench to see if concurrent read performance change?

tabokie:
No significant difference.
8 threads on 40 core, kv fully seperated:
BEFORE:
DB path: [/tmp/rocksdbtest-1000/dbbench]
fillseq      :       3.425 micros/op 291983 ops/sec;   75.7 MB/s
DB path: [/tmp/rocksdbtest-1000/dbbench]
readrandom   :       2.503 micros/op 399520 ops/sec;  103.6 MB/s (10000000 of 10000000 found)
DB path: [/tmp/rocksdbtest-1000/dbbench]
readseq      :       0.463 micros/op 2159710 ops/sec;  560.2 MB/s
AFTER:
DB path: [/tmp/rocksdbtest-1000/dbbench]
fillseq      :       3.361 micros/op 297550 ops/sec;   77.2 MB/s
DB path: [/tmp/rocksdbtest-1000/dbbench]
readrandom   :       2.533 micros/op 394746 ops/sec;  102.4 MB/s (10000000 of 10000000 found)
DB path: [/tmp/rocksdbtest-1000/dbbench]
readseq      :       0.435 micros/op 2300415 ops/sec;  596.7 MB/s

yiwu-arbug:
@tabokie please share your full command line of db_bench.

PR #12, opened 2019-05-22
Add ticker and histrogram metrics

Connor1996:
Because it is a little hard to extend RocksDB statistics, for now, we use BlobDB metrics directly. After making RocksDB statistics easy to extend, I will use Titan own metrics enum then. 

And blob file cache hit/miss is counted with RocksDB block cache
- BLOCK_CACHE_HIT
- BLOCK_CACHE_DATA_HIT
- BLOCK_CACHE_MISS
- BLOCK_CACHE_DATA_MISS

Ticker metrics are: (**some metrics' actual usage is not quite consistent with their names**)
- BLOB_DB_NUM_NEXT, # of Next to get value from blob fo iterator.
- BLOB_DB_NUM_PREV, # of Prev to get value from blob fo iterator.
- BLOB_DB_NUM_KEYS_WRITTEN, # of keys written to blob file.
- BLOB_DB_NUM_KEYS_READ, # of keys read from blob file.
- BLOB_DB_BYTES_WRITTEN, # of bytes (key + value) written due to gc.
- BLOB_DB_BYTES_READ, # of bytes (keys + value) read due to gc.
- BLOB_DB_BLOB_FILE_BYTES_WRITTEN, # of bytes (key + value) written to blob file.
- BLOB_DB_BLOB_FILE_BYTES_READ, # of bytes (key + value) read from blob file.
- BLOB_DB_BLOB_FILE_SYNCED,  # of times blob files being synced.
- BLOB_DB_GC_NUM_FILES, # of blob files being garbage collected.
- BLOB_DB_GC_NUM_NEW_FILES, # of blob files generated by garbage collection.
- BLOB_DB_GC_NUM_KEYS_OVERWRITTEN, # of keys drop by garbage collection because they had been overwritten.
- BLOB_DB_GC_BYTES_OVERWRITTEN, size of bytes drop by garbage collection because they had been overwritten.
- BLOB_DB_GC_NUM_KEYS_RELOCATED, # of keys relocated to new blob file by garbage collection.
- BLOB_DB_GC_BYTES_RELOCATED, size of bytes relocated to new blob file by garbage collection.

Histrogram metrics are:
- BLOB_DB_KEY_SIZE,
- BLOB_DB_VALUE_SIZE,
- BLOB_DB_SEEK_MICROS,
- BLOB_DB_NEXT_MICROS,
- BLOB_DB_PREV_MICROS,
- BLOB_DB_BLOB_FILE_WRITE_MICROS,
- BLOB_DB_BLOB_FILE_READ_MICROS,
- BLOB_DB_BLOB_FILE_SYNC_MICROS,
- BLOB_DB_GC_MICROS

And some gauge metrics will be introduced by property later, like: 
- count of blob file
- count of obsolete file
......


src/db_iter.h
```diff
@@ -76,10 +98,11 @@ class TitanDBIterator : public Iterator {
   }
 
  private:
-  void GetBlobValue() {
+  // returns whether the value is in blob file or not
```
tabokie: This bool return isn't clear enough for me. IMO any meaning other than status is confusing. And the NUM_SEEK seems to refer to attempted seek on blob which requires more clarification. In this context I think a test on `iter->IsBlob()` is enough to RecordTick.

src/blob_gc_job.cc
```diff
@@ -48,6 +52,12 @@ class BlobGCJob::GarbageCollectionWriteCallback : public WriteCallback {
       }
     }
 
+    if (!s.ok()) {
+      RecordTick(stats_, BLOB_DB_BLOB_INDEX_EVICTED_COUNT);
+      RecordTick(stats_, BLOB_DB_BLOB_INDEX_EVICTED_SIZE,
+                 key_.size() + index_entry.size());
```
yiwu-arbug: `BLOB_DB_GC_NUM_KEYS_OVERWRITTEN` and `BLOB_DB_GC_BYTES_OVERWRITTEN` is more suitable I think.

src/blob_gc_job.cc
```diff
@@ -341,7 +360,11 @@ bool BlobGCJob::DiscardEntry(const Slice& key, const BlobIndex& blob_index) {
     fprintf(stderr, "GetImpl err, status:%s\n", s.ToString().c_str());
     abort();
   }
+  RecordTick(stats_, BLOB_DB_BYTES_READ, key.size() + index_entry.size());
   if (s.IsNotFound() || !is_blob_index) {
+    RecordTick(stats_, BLOB_DB_GC_NUM_KEYS_OVERWRITTEN);
```
yiwu-arbug: Both `DoSample()` and `DoRunGC()` call `DiscardEntry()`, and we shouldn't log these counter while sampling. Move the logic to `DoRunGC()`? Same for `BLOB_DB_GC_XXX_RELOCATED` counters below.

src/blob_file_builder.cc
```diff
@@ -34,7 +39,9 @@ Status BlobFileBuilder::Finish() {
 
   status_ = file_->Append(buffer);
   if (ok()) {
+    StopWatch sync_sw(env_, stats_, BLOB_DB_BLOB_FILE_SYNC_MICROS);
```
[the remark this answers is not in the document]
Connor1996: Sync is done in the `BatchFinishFiles`, oh I should record metrics there.

src/blob_file_cache.cc
```diff
@@ -59,8 +60,13 @@ Status BlobFileCache::FindFile(uint64_t file_number, uint64_t file_size,
   Status s;
   Slice cache_key = EncodeFileNumber(&file_number);
   *handle = cache_->Lookup(cache_key);
-  if (*handle) return s;
-
+  if (*handle) {
+    RecordTick(stats_, BLOCK_CACHE_DATA_HIT);
```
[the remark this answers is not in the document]
Connor1996: nice catch!

src/blob_gc_job.cc
Connor1996: if do it in `DoRunGC()`, we can't distinguish whether it is overwritten or relocated. Maybe add a bool to decide whether need to count metrics.

src/blob_gc_job.cc
```diff
@@ -26,6 +29,7 @@ class BlobGCJob::GarbageCollectionWriteCallback : public WriteCallback {
       fprintf(stderr, "GetImpl err, status:%s\n", s.ToString().c_str());
       abort();
     }
+    RecordTick(stats_, BLOB_DB_BYTES_READ, key_.size() + index_entry.size());
     if (s.IsNotFound()) {
       // Either the key is deleted or updated with a newer version which is
       // inlined in LSM.
```
yiwu-arbug: All the `Status::Busy` cases below should add to `BLOB_DB_GC_NUM_KEYS_OVERWRITTEN`. And the success case can log to `BLOB_DB_GC_NUM_KEYS_RELOCATED`. I think we can move the ticker counts to `RewriteValidKeyToLSM` by checking return code of `Write()`.

src/blob_gc_job.cc
```diff
@@ -353,7 +380,13 @@ bool BlobGCJob::DiscardEntry(const Slice& key, const BlobIndex& blob_index) {
     abort();
   }
 
-  return !(blob_index == other_blob_index);
+  bool relocate = !(blob_index == other_blob_index);
+  if (relocate && metrics) {
+    RecordTick(stats_, BLOB_DB_GC_NUM_KEYS_RELOCATED);
```
yiwu-arbug: Even when `relocate == true`, the blob record can be rejected by write callback in `RewriteValidKeyToLSM()`. We should increase `BLOB_DB_GC_NUM_KEYS_RELOCATED` there, only after write success.

src/table_builder.cc
```diff
@@ -30,20 +30,26 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+  MeasureTime(stats_, BLOB_DB_KEY_SIZE, key.size());
+  MeasureTime(stats_, BLOB_DB_VALUE_SIZE, value.size());
+
   BlobIndex index;
   BlobRecord record;
   record.key = key;
   record.value = value;
   index.file_number = blob_handle_->GetNumber();
   blob_builder_->Add(record, &index.blob_handle);
+  RecordTick(stats_, BLOB_DB_BLOB_FILE_BYTES_WRITTEN, index.blob_handle.size);
```
yiwu-arbug: Move this counter to `BlobFileBuilder`? GC job doesn't use the table builder, only file builder. I want to use this counter to know the total bytes written to blob files.

src/blob_gc_job.cc
Connor1996: the definition of `BLOB_DB_GC_NUM_KEYS_RELOCATED` is the number of keys dropped by garbage collection because they had been relocated to new blob file. It doesn't indicate the number of keys that have rewrited.  Should I change the definition?

src/table_builder.cc
Connor1996: The bytes written by GC is counted in `BLOB_DB_BYTES_WRITTEN`

Connor1996: I use `BLOB_DB_BLOB_FILE_BYTES_WRITTEN` to indicate the necessary written bytes to blob file and `BLOB_DB_BYTES_WRITTEN` to indicate the extra written bytes introduced by GC.

src/blob_gc_job.cc
Connor1996: ditto

src/blob_gc_job.cc
Connor1996: seems it is no need distinguishing the reason of dropped keys. I will change the definition.

PR #13, opened 2019-05-23
Add thread_safety_test

tabokie:
VersionSet is currently synchronized using Mutex from TitanDBImpl. This approach is inefficient and incurs unnecessary trouble to reference VersionSet outside of DB.
This PR fix this issue by add seperate mutex for VersionSet to synchronize concurrent access to its internal data structure.

### DIFF
--- a/src/thread_safety_test.cc
+++ b/src/thread_safety_test.cc
@@ -174,14 +174,10 @@ TEST_F(TitanThreadSafetyTest, Basic) {
     {
       MutexLock l(&mutex_);
       handles[name] = handle;
-      ref_count[name] = 0;
+      ref_count[name] = job_count;
     }
     for (uint32_t job = 0; job < job_count; job++) {
       threads.emplace_back([&, job, handle, name] {
-        {
-          MutexLock l(&mutex_);
-          ref_count[name]++;
-        }
         for (uint32_t k = 0; k < param_.repeat; k++) {
           jobs[job](handle);
         }
--- a/src/thread_safety_test.cc
+++ b/src/thread_safety_test.cc
@@ -1,7 +1,7 @@
-#include <vector>
 #include <inttypes.h>
 #include <options/cf_options.h>
 #include <map>
+#include <vector>
 
 #include "db_impl.h"
 #include "db_iter.h"
@@ -24,7 +24,7 @@ struct ThreadParam {
 
   uint32_t concurrency{4};
 
-  uint32_t repeat{30};
+  uint32_t repeat{10};
 
   // when set true, faster worker will run extra jobs util slowest
   // worker finishes to maximize race condition.
@@ -70,19 +70,13 @@ class TitanThreadSafetyTest : public testing::Test {
     ASSERT_OK(db_impl_->TEST_StartGC(handle->GetID()));
   }
 
-  void Flush(ColumnFamilyHandle* handle) {
-    FlushOptions fopts;
-    ASSERT_OK(db_->Flush(fopts, handle));
-  }
-
-  void PutMap(uint64_t k, std::map<std::string, std::string>& data) {
+  void PutMap(std::map<std::string, std::string>& data, uint64_t k) {
     std::string key = GenKey(k);
     std::string value = GenValue(k);
     data.emplace(key, value);
   }
 
-  void PutCF(ColumnFamilyHandle* handle,
-             uint64_t k) {
+  void PutCF(ColumnFamilyHandle* handle, uint64_t k) {
     WriteOptions wopts;
     std::string key = GenKey(k);
     std::string value = GenValue(k);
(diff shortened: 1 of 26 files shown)


tabokie:
PTAL @yiwu-arbug

src/version_set.h
```diff
@@ -86,6 +81,10 @@ class VersionSet {
 
   Status Apply(VersionEdit* edit);
 
+  Status LogAndApplyLocked(VersionEdit* edit);
```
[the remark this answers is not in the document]
Connor1996: add a mutex held assert.

tabokie:
Should be faster.

tabokie:
mutex reduction around BlobGCJob could be premature. PTAL @yiwu-arbug

src/version_set.h
```diff
@@ -86,6 +81,11 @@ class VersionSet {
 
   Status Apply(VersionEdit* edit);
 
+  // REQUIRES: write lock is held
+  Status LogAndApplyLocked(VersionEdit* edit);
+
+  port::RWMutex mutex_;
```
yiwu-arbug: I prefer mutex instead of rwlock (i.e. RWMutex) unless we have some benchmark here. My understanding is rwlock is good only for when you want to hold (read) lock for relatively longer time and rwlock overhead can be ignore. My experience with rwlock is it rarely outperform mutex in most of the cases.

src/version_set.cc
```diff
@@ -10,7 +10,8 @@ namespace titandb {
 const size_t kMaxFileCacheSize = 1024 * 1024;
 
 VersionSet::VersionSet(const TitanDBOptions& options)
-    : dirname_(options.dirname),
+    : mutex_(),
```
yiwu-arbug: nit: don't need to be in init list when init without param.

src/version_set.cc
```diff
@@ -254,6 +261,7 @@ Status VersionSet::Apply(VersionEdit* edit) {
 
 void VersionSet::AddColumnFamilies(
     const std::map<uint32_t, TitanCFOptions>& column_families) {
+  ReadLock l(&mutex_);
```
yiwu-arbug: should hold write lock (if we are to use rwlock). This function modifies `column_families_`.

src/version_set.cc
```diff
@@ -266,6 +274,7 @@ void VersionSet::AddColumnFamilies(
 Status VersionSet::DropColumnFamilies(
     const std::vector<uint32_t>& column_families,
     SequenceNumber obsolete_sequence) {
+  ReadLock l(&mutex_);
```
yiwu-arbug: should hold write lock (if we are to use rwlock). This function modifies `obsolete_columns_`.

src/version_set.h
tabokie: noted

src/version_set.cc
tabokie: pardon my mistake. So vset is not read-intensive-like any more.

src/thread_safety_test.cc
```diff
@@ -0,0 +1,232 @@
...
+    options_.dirname = dbname_ + "/titandb";
+    options_.create_if_missing = true;
+    options_.min_blob_size = 32;
+    options_.min_gc_batch_size = 1;
+    options_.blob_file_compression = CompressionType::kLZ4Compression;
+    DeleteDir(env_, options_.dirname);
+    DeleteDir(env_, dbname_);
+  }
+
+  ~TitanThreadSafetyTest() { Close(); }
```
yiwu-arbug: minor: delete DB directory after close?

src/thread_safety_test.cc
```diff
@@ -0,0 +1,220 @@
...
+    {
+      MutexLock l(&mutex_);
+      handles[name] = handle;
+      ref_count[name] = 0;
+    }
+    for (uint32_t job = 0; job < job_count; job++) {
+      threads.emplace_back([&, job, handle, name] {
+        {
+          MutexLock l(&mutex_);
+          ref_count[name]++;
```
yiwu-arbug: Seems it is possible that some threads may start after the others finished and dropped the CF. Not sure if it cause some of the TSAN error. Seems to me you can simply set ref_count[name] = job_count before the loop.

tabokie: good point

yiwu-arbug:
One of the TSAN failure is from race condition on
* `VersionSet::LogAndApply()` will recompute GC score by calling `BlobStorage::ComputeGCScore()`
* A GC job may be running from another thread and access `BlobStorage::gc_score_`.

Let's change the rwlock in `BlobStorage` to mutex and hold mutex lock the whole time in `BlobStorage::ComputeGCScore` and `BlobStorage::gc_score()`.

tabokie:
Can't reproduce the failed test on my end. PTAL @yiwu-arbug

yiwu-arbug:
Since now #11 don't depend on this one, how about we keep the thread_safety_test and abandon the rest of changes?

tabokie:
ci passed, PTAL @yiwu-arbug

Connor1996:
So we don't add separate mutex for VersionSet now?

yiwu-arbug:
@Connor1996 I prefer to keep our code stable, as we approach 3.0 GA.

Connor1996:
@tabokie Please update PR description

PR #15, opened 2019-05-24
Add Titan internal stats and metrics for blob file and live value

tabokie:
Titan InternalStats provides a global counter that can decrement as well as increment. Thread safety is guaranteed by plain atomic variable.

Add three metrics with the use of InternalStats:

* LIVE_BLOB_SIZE: total size of value referenced from LSM tree.
* NUM_BLOB_FILE: number of blob file.
* BLOB_FILE_SIZE: total size of blob file.

Also, to save parameter passing, introduce TitanStats as a wrapper around RocksDB Statistics and Titan's internal stats.

### DIFF
--- a/src/blob_storage.cc
+++ b/src/blob_storage.cc
@@ -25,7 +25,7 @@ Status BlobStorage::NewPrefetcher(uint64_t file_number,
 }
 
 std::weak_ptr<BlobFileMeta> BlobStorage::FindFile(uint64_t file_number) const {
-  ReadLock rl(&mutex_);
+  MutexLock l(&mutex_);
   auto it = files_.find(file_number);
   if (it != files_.end()) {
     assert(file_number == it->second->file_number());
@@ -36,14 +36,14 @@ std::weak_ptr<BlobFileMeta> BlobStorage::FindFile(uint64_t file_number) const {
 
 void BlobStorage::ExportBlobFiles(
     std::map<uint64_t, std::weak_ptr<BlobFileMeta>>& ret) const {
-  ReadLock rl(&mutex_);
+  MutexLock l(&mutex_);
   for (auto& kv : files_) {
     ret.emplace(kv.first, std::weak_ptr<BlobFileMeta>(kv.second));
   }
 }
 
 void BlobStorage::AddBlobFile(std::shared_ptr<BlobFileMeta>& file) {
-  WriteLock wl(&mutex_);
+  MutexLock l(&mutex_);
   files_.emplace(std::make_pair(file->file_number(), file));
   AddStats(stats_, cf_id_, TitanInternalStats::LIVE_BLOB_FILE_SIZE,
            file->file_size());
@@ -52,7 +52,7 @@ void BlobStorage::AddBlobFile(std::shared_ptr<BlobFileMeta>& file) {
 
 void BlobStorage::MarkFileObsolete(std::shared_ptr<BlobFileMeta> file,
                                    SequenceNumber obsolete_sequence) {
-  WriteLock wl(&mutex_);
+  MutexLock l(&mutex_);
   obsolete_files_.push_back(
       std::make_pair(file->file_number(), obsolete_sequence));
   file->FileStateTransit(BlobFileMeta::FileEvent::kDelete);
@@ -68,7 +68,7 @@ void BlobStorage::MarkFileObsolete(std::shared_ptr<BlobFileMeta> file,
 
 void BlobStorage::GetObsoleteFiles(std::vector<std::string>* obsolete_files,
                                    SequenceNumber oldest_sequence) {
-  WriteLock wl(&mutex_);
+  MutexLock l(&mutex_);
 
   uint32_t file_dropped = 0;
   uint64_t file_dropped_size = 0;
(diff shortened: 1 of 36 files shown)


src/titan_stats.h
```diff
@@ -0,0 +1,136 @@
...
+    if (p == internal_stats_.end()) {
+      return nullptr;
+    } else {
+      return p->second.get();
+    }
+  }
+  TitanInternalStats* internal_stats() { return internal_stats(default_cf_); }
+
+ private:
+  Statistics* stats_ = nullptr;
```
yiwu-arbug: Let's not combine `Statistics` with `TitanStats`. From TiKV for both plain RocksDB and Titan, we pass `Statistics` through `DBOptions`. With the change here, we need to tell apart whether we are using plain RocksDB and Titan, and pass `Statistics` object differently, which make the interface more cumbersome.

tabokie: Remove TitanStats from options to clear the trouble.

src/titan_stats.h
```diff
@@ -0,0 +1,136 @@
...
+  Statistics* statistics() { return stats_; }
+  TitanInternalStats* internal_stats(uint32_t cf_id) {
+    auto p = internal_stats_.find(cf_id);
+    if (p == internal_stats_.end()) {
+      return nullptr;
+    } else {
+      return p->second.get();
+    }
+  }
+  TitanInternalStats* internal_stats() { return internal_stats(default_cf_); }
```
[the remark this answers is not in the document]
tabokie: It does support. This is a helper getter for those modules without knowledge of cf_id.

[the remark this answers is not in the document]
yiwu-arbug: All properties we added so far are CF specific (e.g. "num-blob-file" should return number of blob files for a specific CF, and so on so forth). `internal_stats()` and all overload of `AddStats`, `SubStats` and `ResetStats` without CF id can be removed.

src/blob_storage.cc
```diff
@@ -45,6 +45,11 @@ void BlobStorage::ExportBlobFiles(
 void BlobStorage::AddBlobFile(std::shared_ptr<BlobFileMeta>& file) {
   WriteLock wl(&mutex_);
   files_.emplace(std::make_pair(file->file_number(), file));
+  AddStats(stats_, cf_id_, TitanInternalStats::LIVE_BLOB_SIZE,
```
yiwu-arbug: Will all these properties be reconstructed on DB reopen?

src/db_impl.cc
```diff
@@ -603,18 +631,23 @@ void TitanDBImpl::OnCompactionCompleted(
...
+      delta += -bfs.second;
       auto file = bs->FindFile(bfs.first).lock();
       if (!file) {
         // file has been gc out
         continue;
       }
       file->AddDiscardableSize(static_cast<uint64_t>(-bfs.second));
     }
+    SubStats(stats_.get(), compaction_job_info.cf_id,
+             TitanInternalStats::LIVE_BLOB_SIZE, delta);
```
yiwu-arbug: should only update stat if the file is not obsolete

src/db_impl.cc
```diff
@@ -603,18 +631,23 @@ void TitanDBImpl::OnCompactionCompleted(
...
     }
 
+    uint64_t delta = 0;
     for (const auto& bfs : blob_files_size) {
+      delta += bfs.second;
       // blob file size < 0 means discardable size > 0
       if (bfs.second >= 0) {
         continue;
       }
+      delta += -bfs.second;
```
yiwu-arbug: The logic around `delta` doesn't look right to me. Line 636 should be remove I think.

tabokie: (how careless am i)

src/titan_stats.h
```diff
@@ -0,0 +1,168 @@
...
+    return false;
+  }
+
+ private:
+  static const std::unordered_map<std::string, TitanInternalStats::StatsType>
+      stats_type_string_map;
+  std::atomic<uint64_t> stats_[INTERNAL_STATS_ENUM_MAX];
+};
+
+class TitanStats {
```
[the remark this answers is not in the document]
tabokie: Seperating them is ok. I just don't want to make room for ugly interface, like passing a rocksdb::stats to A while passing a titan::stats to B.

src/blob_storage.cc
tabokie: I think they are. During `Apply` in `VersionSet::Recover`.

src/titan_stats.h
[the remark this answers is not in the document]
tabokie: If `statistics(stats_)` is too tedious, we can patch `RecordTick` instead. Your call.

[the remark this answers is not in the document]
yiwu-arbug: ok. let's replace `RecordTick` with another helper method to do `RecordTick(stats_->statistics(),...)`.

src/blob_storage.cc
yiwu-arbug: Hmm, it seems to me `VersionSet::Apply` doesn't reset `discardable_size`, so `live-blob-size` will be incorrect anyway. Let's address the issue separately.

src/blob_storage.cc
```diff
@@ -45,6 +45,11 @@ void BlobStorage::ExportBlobFiles(
...
+           file->file_size());
```
Connor1996: seems it doesn't follow the definition `total blob value size referenced by LSM tree`. `file_size` includes header + footer + key

tabokie:
Latest fix should resolve build failure on master branch, PTAL @yiwu-arbug

PR #17, opened 2019-06-04
Fix version set log

Connor1996:
- the info_log of `db_options` passed to `version_set` is null, so create `version_set` until we set the info_log
- add output file log for gc job


src/db_impl.cc
```diff
@@ -187,12 +186,13 @@ Status TitanDBImpl::Open(const std::vector<TitanCFDescriptor>& descs,
   }
   if (!s.ok()) return s;
 
-  s = vset_->Open(column_families);
-  if (!s.ok()) return s;
-
   // Add EventListener to collect statistics for GC
   db_options_.listeners.emplace_back(std::make_shared<BaseDbListener>(this));
 
+  vset_.reset(new VersionSet(db_options_));
```
yiwu-arbug: add comment saying this is to set info log?

src/blob_gc_job.cc
```diff
@@ -423,12 +424,20 @@ Status BlobGCJob::InstallOutputBlobFiles() {
...
+
+      if (!tmp.empty()) {
+        tmp.append(" ");
+      }
+      tmp.append(std::to_string(file->file_number()));
+
       blob_gc_->AddOutputFile(file.get());
       files.emplace_back(std::make_pair(file, std::move(builder.first)));
     }
+    ROCKS_LOG_BUFFER(log_buffer_, " output[%s]", tmp.c_str());
```
yiwu-arbug: Also print CF id like the "Titan GC candidates xxx" log?

PR #18, opened 2019-06-05
ignore missing blob file for iterator

Connor1996:
- add key-only iter option
- reproduce missing blob file when iterating
- fix an issue about sample which affecting reproducing

### DIFF
--- a/include/titan/db.h
+++ b/include/titan/db.h
@@ -77,31 +77,6 @@ class TitanDB : public StackableDB {
   Status DestroyColumnFamilyHandle(ColumnFamilyHandle* column_family) override =
       0;
 
-  using StackableDB::Get;
-  Status Get(const ReadOptions& options,
-                     ColumnFamilyHandle* column_family, const Slice& key,
-                     PinnableSlice* value) override {
-        
-    return Get(TitanReadOptions(options), column_family, key, value);
-  }
-  virtual Status Get(const TitanReadOptions& options,
-                     ColumnFamilyHandle* column_family, const Slice& key,
-                     PinnableSlice* value) = 0;
-
-  using StackableDB::MultiGet;
-  std::vector<Status> MultiGet(
-      const ReadOptions& options,
-      const std::vector<ColumnFamilyHandle*>& column_family,
-      const std::vector<Slice>& keys,
-      std::vector<std::string>* values) override {
-    return MultiGet(TitanReadOptions(options), column_family, keys, values);
-  }
-  virtual std::vector<Status> MultiGet(
-      const TitanReadOptions& options,
-      const std::vector<ColumnFamilyHandle*>& column_family,
-      const std::vector<Slice>& keys,
-      std::vector<std::string>* values) = 0;
-
   using StackableDB::NewIterator;
   Iterator* NewIterator(const ReadOptions& opts,
                                 ColumnFamilyHandle* column_family) override {
--- a/include/titan/db.h
+++ b/include/titan/db.h
@@ -77,6 +77,51 @@ class TitanDB : public StackableDB {
   Status DestroyColumnFamilyHandle(ColumnFamilyHandle* column_family) override =
       0;
 
+  using StackableDB::Get;
+  Status Get(const ReadOptions& options,
+                     ColumnFamilyHandle* column_family, const Slice& key,
+                     PinnableSlice* value) override {
+        
+    return Get(TitanReadOptions(options), column_family, key, value);
+  }
+  virtual Status Get(const TitanReadOptions& options,
+                     ColumnFamilyHandle* column_family, const Slice& key,
+                     PinnableSlice* value) = 0;
+
+  using StackableDB::MultiGet;
+  std::vector<Status> MultiGet(
+      const ReadOptions& options,
+      const std::vector<ColumnFamilyHandle*>& column_family,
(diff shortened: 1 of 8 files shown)


src/db_iter.h
```diff
@@ -96,17 +109,23 @@ class TitanDBIterator : public Iterator {
       std::unique_ptr<BlobFilePrefetcher> prefetcher;
       status_ = storage_->NewPrefetcher(index.file_number, &prefetcher);
       if (status_.IsCorruption()) {
+        // If use `DeleteFilesInRanges`, we may encounter this failure,	
+        // because `DeleteFilesInRanges` may expose an old key which
```
tabokie: "whose"

src/db_iter.h
```diff
@@ -96,17 +109,23 @@ class TitanDBIterator : public Iterator {
       std::unique_ptr<BlobFilePrefetcher> prefetcher;
       status_ = storage_->NewPrefetcher(index.file_number, &prefetcher);
       if (status_.IsCorruption()) {
```
tabokie: Seems vulnerable to me. Any other unexpected and unresolved corruption could cause an infinite loop. Should work as a temporary patch though.

Connor1996: why infinite loop? Line 93 will break the loop in final

src/db_iter.h
```diff
@@ -96,17 +109,23 @@ class TitanDBIterator : public Iterator {
...
+        // If use `DeleteFilesInRanges`, we may encounter this failure,	
+        // because `DeleteFilesInRanges` may expose an old key which	
+        // corresponding blob file has already been GCed out, so we	
+        // cannot abort here.
         fprintf(stderr,
                 "key:%s GetBlobValue err:%s with sequence number:%" PRIu64 "\n",
                 iter_->key().ToString(true).c_str(), status_.ToString().c_str(),
                 options_.snapshot->GetSequenceNumber());
+        return false;
```
yiwu-arbug: change the `fprintf` to print to info log? And I think we shall print only once for each iterator.

src/db_iter.h
tabokie: No loop then. But we could do better than catching a generic error in future.

Connor1996:
PTAL again @yiwu-arbug @tabokie

src/blob_gc_job.cc
```diff
@@ -161,13 +161,15 @@ bool BlobGCJob::DoSample(const BlobFileMeta* file) {
...
 
   // TODO: add do sample count metrics
-
+  auto records_size = file->file_size() - BlobFileHeader::kEncodedLength -
+                      BlobFileFooter::kEncodedLength;
   Status s;
   uint64_t sample_size_window = static_cast<uint64_t>(
-      file->file_size() * blob_gc_->titan_cf_options().sample_file_size_ratio);
-  Random64 random64(file->file_size());
+      records_size * blob_gc_->titan_cf_options().sample_file_size_ratio);
```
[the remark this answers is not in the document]
Connor1996: for sampling, not include header and footer when calculating the discardable ratio.

Connor1996:
PTAL again @yiwu-arbug @tabokie

src/blob_gc_job_test.cc
```diff
@@ -84,14 +86,23 @@ class BlobGCJobTest : public testing::Test {
...
+
   void DestroyDB() {
     Status s __attribute__((__unused__)) = db_->Close();
     assert(s.ok());
     delete db_;
     db_ = nullptr;
   }
 
-  void RunGC() {
+  void RunGC(bool expected = false) {
```
tabokie: Seems a lot similiar with `TEST_StartGC`. Maybe remove and update that function?

Connor1996: let's just keep that in this PR. IMO `TEST_StartGC` is still too similar to `BackgroundGC`, we can merge them together.

include/titan/db.h
```diff
@@ -77,6 +77,26 @@ class TitanDB : public StackableDB {
...
+
+  using StackableDB::NewIterators;
+  Status NewIterators(
+      const ReadOptions& options,
+      const std::vector<ColumnFamilyHandle*>& column_families,
+      std::vector<Iterator*>* iterators) override {
+    return NewIterators(TitanReadOptions(options), column_families, iterators);
+  }
+  virtual Status NewIterators(
+      const TitanReadOptions& options,
```
yiwu-arbug: Shall we split the struct into two (e.g. `NewIterators(rocksdb::ReadOptions, TitanReadOptions, ...)`)? That's consistent with what we do on rust-rocksdb side.

src/db_iter.h
```diff
@@ -98,14 +98,14 @@ class TitanDBIterator : public Iterator {
   }
 
   Slice value() const override {
-    assert(Valid());
+    assert(Valid() && !options_.key_only);
```
yiwu-arbug: return `Slice()` if `key_only=true`?

Connor1996: should we abort or set the status to `not ok`?

yiwu-arbug: Just return `Slice()`? I think it's ok to call `value()` in case `key_only=true`.

include/titan/db.h
yiwu-arbug: discussed offline. Let's keep it the current way.

PR #19, opened 2019-06-10
Introduce edit collector

Connor1996:
Introduce edit collector to apply the VersionEdits in batch


src/version_edit.h
```diff
@@ -42,5 +45,52 @@ class VersionEdit {
...
+      collector.DeleteFile(file.first, file.second);
+      if (!s.ok()) return s;
+    }
+    return Status::OK();
+  }
+
+ private:
+  friend class VersionSet;
+
+  class Collector {
```
yiwu-arbug: s/Collector/CFEditCollector ?

src/version_edit.h
```diff
@@ -42,5 +45,52 @@ class VersionEdit {
...
+   public:
+    Collector() {}
```
yiwu-arbug: nit: this can be remove.

src/version_edit.h
```diff
@@ -42,5 +45,52 @@ class VersionEdit {
...
+
+    Status AddFile(const std::shared_ptr<BlobFileMeta>& file) {
+      auto number = file->file_number();
+      if (added_files_.find(number) != added_files_.end()) {
```
yiwu-arbug: nit: could be just `added_files_.count(number) > 0`.

src/version_edit.h
```diff
@@ -42,5 +45,52 @@ class VersionEdit {
...
+        return Status::Corruption("blob file has been added before");
+      }
+      added_files_.emplace(number, file);
+      return Status::OK();
+    }
+
+    Status DeleteFile(uint64_t number, SequenceNumber obsolete_sequence) {
+      if (!added_files_.erase(number)) {
+        deleted_files_.emplace_back(std::make_pair(number, obsolete_sequence));
```
yiwu-arbug: I'm a bit paranoid, but let's also check if the file is deleted before.

src/version_edit.h
```diff
@@ -42,5 +45,52 @@ class VersionEdit {
...
+      auto number = file->file_number();
+      if (added_files_.find(number) != added_files_.end()) {
+        return Status::Corruption("blob file has been added before");
+      }
+      added_files_.emplace(number, file);
+      return Status::OK();
+    }
+
+    Status DeleteFile(uint64_t number, SequenceNumber obsolete_sequence) {
+      if (!added_files_.erase(number)) {
```
yiwu-arbug: let's collect the set of file added and deleted, and sanity check the file don't appear later on.

src/version_edit.h
```diff
@@ -33,6 +35,7 @@ class VersionEdit {
 
  private:
   friend class VersionSet;
+  friend class EditCollector;
 
   bool has_next_file_number_{false};
   uint64_t next_file_number_{0};
```
yiwu-arbug: I think the logic is cleaner if `EditCollector` also handle `next_file_number_`.

src/version_set.cc
```diff
@@ -204,52 +207,61 @@ Status VersionSet::LogAndApply(VersionEdit* edit) {
...
-    auto number = file.first;
-    auto blob_it = files.find(number);
-    if (blob_it == files.end()) {
-      fprintf(stderr, "blob file %" PRIu64 " doesn't exist before\n", number);
-      abort();
-    } else if (blob_it->second->is_obsolete()) {
-      fprintf(stderr, "blob file %" PRIu64 " has been deleted before\n",
-              number);
-      abort();
+Status VersionSet::Apply(EditCollector* collector) {
```
yiwu-arbug: Change this to `EditCollector&`? RocksDB convention is to use reference for input param, and pointer for output param.

src/version_edit.h
```diff
@@ -33,6 +35,7 @@ class VersionEdit {
 
  private:
   friend class VersionSet;
+  friend class EditCollector;
```
yiwu-arbug: Adding some getters is better than make classes friend each other.

src/version_set.cc
```diff
@@ -204,52 +207,61 @@ Status VersionSet::LogAndApply(VersionEdit* edit) {
...
       }
-      abort();
+      it->second->AddBlobFile(file.second);
     }
-    it->second->AddBlobFile(file);
-  }
 
-  it->second->ComputeGCScore();
+    it->second->ComputeGCScore();
+  }
```
yiwu-arbug: let's also sanity check those files deleted after added in `EditCollector` does not exist in `VersionSet`.

src/version_edit.h
Connor1996: What do you mean? `EditCollector` checks whether `next_file_number_` is monotonic increasing?

Connor1996: Oh, I got it.

src/edit_collector.h
```diff
@@ -0,0 +1,135 @@
...
+      s = collector.AddFile(file);
+      if (!s.ok()) return s;
+    }
+    for (auto& file : edit.deleted_files_) {
+      collector.DeleteFile(file.first, file.second);
+      if (!s.ok()) return s;
+    }
+
+    if (edit.has_next_file_number_) {
+      assert(edit.next_file_number_ >= next_file_number_);
```
yiwu-arbug: return Corruption if this check fail?

PR #22, opened 2019-06-14
Patch test for blob-run-mode

tabokie:
Newly introduced fallback mode of blob-run-mode is causing `Missing Blob File` fatal error, thus this PR adds more multi-threaded test around blob-run-mode setting. So far no bug is caught by unit test, Any idea on more solid test is welcome.

misc: fix file permission setting, fix SetOptions test in db_test.


Connor1996:
How it cause `Missing blob file`?

tabokie:
> How it causes `Missing blob file`?

For now use `tikv-ctl` to enter `fallback` mode while inserting into tikv will cause `Missing Blob File`.

yiwu-arbug:
I reproduced the issue and found out why. The line here need to ignore files that has been GC, otherwise it will fail compaction. https://github.com/pingcap/titan/blob/master/src/table_builder.cc#L32

tabokie:
Great! I will add GC invocation to strengthen unittest later.

yiwu-arbug:
issue fixed by #30

PR #24, opened 2019-06-17
misc: unified scope for unique_ptr

zhouyuan:


yiwu-arbug:
@zhouyuan Thanks for contribution! Are you testing Titan?

zhouyuan:
@yiwu-arbug Hi, trying to do some basic tests to check the compaction overhead here. I understand Titan is built for TiKV/TiDB. My interest is to investigate how does the performance looks like with different values(100B ~1KB) - it's smaller but common for metadada case mostly.  
FWIW RocksDB works very good as the metadata store of Ceph except the latency spike brought by compaction is annoying. Ceph community is also investigating on TRocksDB. 
https://www.spinics.net/lists/ceph-users/msg52297.html

Thanks for the great work!

thanks, -yuan

yiwu-arbug:
@zhouyuan thanks for the pointers, especially the TRocksDB one.

PR #26, opened 2019-06-20
Eliminate use of abort()

yiwu-arbug:
Summary:
Current implementation call `abort()` in cases when it cannot handle the error. This patch eliminate the use of `abort()` by either return proper error code, or log error to error log and move on.

For critical errors, what rocksdb do is to turn the DB into read only mode, and wait for user to handle the error (instead of crashing). I'll send a patch to do the same later.


src/db_iter.h
```diff
@@ -118,22 +121,16 @@ class TitanDBIterator : public Iterator {
...
     if (it == files_.end()) {
       std::unique_ptr<BlobFilePrefetcher> prefetcher;
       status_ = storage_->NewPrefetcher(index.file_number, &prefetcher);
-      if (status_.IsCorruption()) {
-        fprintf(stderr,
-                "key:%s GetBlobValue err:%s with sequence number:%" PRIu64 "\n",
-                iter_->key().ToString(true).c_str(), status_.ToString().c_str(),
-                options_.snapshot->GetSequenceNumber());
+      if (!status_.ok()) {
+        return;
```
Connor1996: no need to print error log here?

src/blob_gc_job.cc
```diff
@@ -187,12 +190,13 @@ bool BlobGCJob::DoSample(const BlobFileMeta* file) {
...
   if (!iter.status().ok()) {
-    fprintf(stderr,
-            "IterateForPrev faile, file number[%lu] size[%lu] status[%s]\n",
-            static_cast<size_t>(file->file_number()),
-            static_cast<size_t>(file->file_size()),
-            iter.status().ToString().c_str());
-    abort();
+    s = iter.status();
+    ROCKS_LOG_ERROR(db_options_.info_log,
+                    "IterateForPrev faile, file number[%" PRIu64
```
Connor1996:
```suggestion
                    "IterateForPrev failed, file number[%" PRIu64
```

src/db_impl.cc
```diff
@@ -616,8 +621,12 @@ void TitanDBImpl::OnFlushCompleted(const FlushJobInfo& flush_job_info) {
...
     auto blob_storage = vset_->GetBlobStorage(flush_job_info.cf_id).lock();
     if (!blob_storage) {
-      fprintf(stderr, "Column family id:%u Not Found\n", flush_job_info.cf_id);
-      abort();
+      // TODO: Should treat it as background error and make DB read-only.
+      ROCKS_LOG_ERROR(db_options_.info_log,
+                      "OnFlushCompleted[%d]: Column family id: %" PRIu32
+                      " Not Found.",
+                      flush_job_info.job_id, flush_job_info.cf_id);
+      assert(false);
```
Connor1996: missing return

src/db_impl.cc
```diff
@@ -632,19 +641,18 @@ void TitanDBImpl::OnFlushCompleted(const FlushJobInfo& flush_job_info) {
...
   std::set<uint64_t> outputs;
   std::set<uint64_t> inputs;
-  auto calc_bfs = [&compaction_job_info, &blob_files_size, &outputs, &inputs](
-                      const std::vector<std::string>& files, int coefficient,
+  auto calc_bfs = [&](const std::vector<std::string>& files, int coefficient,
                       bool output) {
     for (const auto& file : files) {
       auto tp_iter = compaction_job_info.table_properties.find(file);
       if (tp_iter == compaction_job_info.table_properties.end()) {
-        if (output) {
```
Connor1996: why remove it?

src/db_impl.cc
```diff
@@ -685,15 +699,22 @@ void TitanDBImpl::OnCompactionCompleted(
...
+      auto file = bs->FindFile(file_number).lock();
       if (!file) {
-        fprintf(stderr, "OnCompactionCompleted get file failed\n");
-        abort();
+        // TODO: Should treat it as background error and make DB read-only.
+        ROCKS_LOG_ERROR(
+            db_options_.info_log,
+            "OnCompactionCompleted[%d]: Failed to get file %" PRIu64,
+            compaction_job_info.job_id, file_number);
+        assert(false);
```
Connor1996: missing return

src/blob_file_size_collector.cc
```diff
@@ -72,10 +72,9 @@ Status BlobFileSizeCollector::Finish(UserCollectedProperties* properties) {
   }
 
   std::string res;
-  if (!Encode(blob_files_size_, &res) || res.empty()) {
-    fprintf(stderr, "blob file size collector encode failed\n");
-    abort();
-  }
+  bool ok __attribute__((__unused__)) = Encode(blob_files_size_, &res);
+  assert(ok);
```
Connor1996: we may ignore the error in release mode?

src/db_impl.cc
yiwu-arbug: Seems to me it if fine if compaction output file don't have table property. But let me add a warn level log here.

PR #32, opened 2019-06-27
Add GetTitanOptions() and other options related changes

yiwu-arbug:
Summary:
* Add `GetTitanOptions()` and `GetTitanDBOptions()` API.
* Fix `SetOptions()` fail when `blob_run_mode` is not set.
* Update info log around printing options.

Test Plan:
Updated test.

### DIFF
--- a/src/db_impl.cc
+++ b/src/db_impl.cc
@@ -200,6 +200,10 @@ Status TitanDBImpl::Open(const std::vector<TitanCFDescriptor>& descs,
       // While we need to preserve original table_factory for GetOptions.
       auto& base_table_factory = base_descs[i].options.table_factory;
       assert(base_table_factory != nullptr);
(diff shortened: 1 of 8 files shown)


src/db_impl.cc
```diff
@@ -608,23 +626,42 @@ Status TitanDBImpl::SetOptions(
...
   }
-  return Status::OK();
+  return titan_options;
+}
+
+TitanDBOptions TitanDBImpl::GetTitanDBOptions() const {
+  // Titan db_options_ is not mutable after DB open.
+  TitanDBOptions result = db_options_;
+  *static_cast<DBOptions*>(&result) = db_impl_->GetDBOptions();
+  return db_options_;
```
Connor1996: return result?

yiwu-arbug: good catch!

PR #36, opened 2019-07-10
Trigger next gc if needed after a gc

GlitterIsMe:
1. **Trigger next GC if needed after a GC if needed**: When pick file for gc, gc_picker will keep reading BlobFileMeta in BlobStorage and record the total file size (next_gc_size) that can be GC in the remain files, when the next_gc_size exceeds min_gc_batch_size, it maybe necessary to trigger next GC when this GC is over. The result of this check will be carried by BlobGC and can be accessed by BlobGC::tigger_next().
2. **Take the size of output file into consideration**:  I add a min_output_threshold which is the minimum value of merge_file_threshold and blob_file_target_size / 4 by default. Only when the batch size exceeds max_gc_batch_size and estimate_output_size(which is total_batch_size - total_discardable_size) is larger than min_output_threshold can the picker stop picking files.


src/blob_storage.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
-
+#ifndef __STDC_FORMAT_MACROS
+#define  __STDC_FORMAT_MACROS
```
Connor1996: extra space

src/blob_gc_picker.cc
```diff
@@ -15,6 +15,8 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
   std::vector<BlobFileMeta*> blob_files;
 
   uint64_t batch_size = 0;
+  uint64_t min_output_threshold = std::max(cf_options_.merge_small_file_threshold, cf_options_.blob_file_target_size / 4);
+  uint64_t estimate_output_size = 0;
   //  ROCKS_LOG_INFO(db_options_.info_log, "blob file num:%lu gc score:%lu",
```
Connor1996: maybe add some debug logs here?

src/blob_gc_picker.cc
```diff
@@ -40,7 +42,8 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
     blob_files.push_back(blob_file.get());
 
     batch_size += blob_file->file_size();
-    if (batch_size >= cf_options_.max_gc_batch_size) break;
+    estimate_output_size += (blob_file->file_size() - blob_file->discardable_size());
+    if (batch_size >= cf_options_.max_gc_batch_size && estimate_output_size >= min_output_threshold) break;
```
Connor1996: can we just readahead some files here to decide whether it needs to do next GC? I think calculating `PenddingGCSize` every GC is wasteful.

GlitterIsMe: OK, maybe a more convenient way

src/db_impl_gc.cc
```diff
@@ -44,6 +43,7 @@ void TitanDBImpl::BackgroundCallGC() {
     }
 
     bg_gc_scheduled_--;
+    MaybeScheduleGC();
```
Connor1996: why do schedule here?

src/db_impl_gc.cc
```diff
@@ -14,13 +14,12 @@ void TitanDBImpl::MaybeScheduleGC() {
...
   if (shuting_down_.load(std::memory_order_acquire)) return;
 
-  if (bg_gc_scheduled_.load(std::memory_order_acquire) >=
-      db_options_.max_background_gc)
-    return;
-
-  bg_gc_scheduled_.fetch_add(1, std::memory_order_release);
-
-  env_->Schedule(&TitanDBImpl::BGWorkGC, this, Env::Priority::LOW, this);
+  while (unscheduled_gc_ > 0 &&
```
Connor1996: please extract the change to the other PR

src/blob_gc_picker_test.cc
```diff
@@ -73,6 +73,39 @@ TEST_F(BlobGCPickerTest, BeingGC) {
...
 }
 
+TEST_F(BlobGCPickerTest, PickMoreFile) {
+  TitanDBOptions titan_db_options;
+  TitanCFOptions titan_cf_options;
+  titan_cf_options.max_gc_batch_size = 1 << 30;
+  titan_cf_options.blob_file_target_size = 256 << 20;
+  // min_output_threshold = 64MB
+  NewBlobStorageAndPicker(titan_db_options, titan_cf_options);
+  AddBlobFile(1U, 1U << 30, 1000U << 20);  // valid_size = 24MB
```
Connor1996: please add more cases, like a lot of files nearly all invalid.

src/blob_gc_picker_test.cc
```diff
@@ -73,6 +73,39 @@ TEST_F(BlobGCPickerTest, BeingGC) {
...
+  titan_cf_options.blob_file_target_size = 256 << 20;
+  NewBlobStorageAndPicker(titan_db_options, titan_cf_options);
+  AddBlobFile(1U, 1U << 30, 1000U << 20);  // valid_size = 24MB
+  AddBlobFile(2U, 1U << 30, 512U << 20);   // valid_size = 512MB
+  AddBlobFile(3U, 1U << 30, 512U << 20);   // valid_size = 512MB
+  AddBlobFile(4U, 1U << 30, 512U << 20);   // valid_size = 512MB
+  UpdateBlobStorage();
+  auto blob_gc = basic_blob_gc_picker_->PickBlobGC(blob_storage_.get());
+  ASSERT_TRUE(blob_gc != nullptr);
+  ASSERT_EQ(blob_gc->trigger_next(), true);
```
Connor1996: add case that `trigger_next()` is false

Connor1996:
Please add PR description

src/db_impl_gc.cc
```diff
@@ -81,6 +81,14 @@ Status TitanDBImpl::BackgroundGC(LogBuffer* log_buffer) {
         assert(column_family_id == cfh->GetID());
         blob_gc->SetColumnFamily(cfh.get());
       }
+
+      if (blob_gc->trigger_next()) {
```
Connor1996: to avoid overwhelming the queue, maybe not trigger next if there are already a lot of jobs to do?

src/db_impl_gc.cc
GlitterIsMe: this point is at the end of a GC, if there are some tasks in the queue, they will be scheduled, or if there is no task or the bg_thread run out nothing will happen.

src/blob_gc_picker.cc
```diff
@@ -37,17 +44,42 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
...
-    blob_files.push_back(blob_file.get());
 
-    batch_size += blob_file->file_size();
-    if (batch_size >= cf_options_.max_gc_batch_size) break;
+    if (!stop_picking) {
+      blob_files.push_back(blob_file.get());
+      batch_size += blob_file->file_size();
+      estimate_output_size +=
+          (blob_file->file_size() - blob_file->discardable_size());
+      if (batch_size >= cf_options_.max_gc_batch_size &&
```
Connor1996: it should be `||` ?

src/blob_gc_picker.cc
```diff
@@ -37,17 +44,42 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
...
+          estimate_output_size >= min_output_threshold) {
+        // stop pick file for this gc, but still check file for whether need
+        // trigger gc after this
+        stop_picking = true;
+      }
+    } else {
+      if (blob_file->GetDiscardableRatio() >=
```
Connor1996: if not, break directly

src/blob_gc_picker.cc
```diff
@@ -37,17 +44,42 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
...
+                         "remain more than %" PRIu64
+                         " bytes to be gc and trigger after this gc",
+                         next_gc_size);
+          break;
+        }
+      }
+    }
   }
-
+  ROCKS_LOG_INFO(db_options_.info_log,
```
Connor1996:
```suggestion
  ROCKS_LOG_DEBUG(db_options_.info_log,
```

src/blob_gc_picker_test.cc
```diff
@@ -73,6 +78,84 @@ TEST_F(BlobGCPickerTest, BeingGC) {
...
+}
+
+TEST_F(BlobGCPickerTest, NotTriggerNext) {
```
Connor1996: just merge it with the above case

src/blob_storage.h
```diff
@@ -1,7 +1,8 @@
 #pragma once
-
+#ifndef __STDC_FORMAT_MACROS
```
Connor1996: seems it should in `blob_gc_picker` instead of here

src/db_impl.h
```diff
@@ -120,6 +120,7 @@ class TitanDBImpl : public TitanDB {
 
   // REQUIRE: mutex_ held
   void AddToGCQueue(uint32_t column_family_id) {
+    unscheduled_gc_++;
```
Connor1996: seems useless, it can be deduced from `gc_queue_`

src/db_impl_gc.cc
```diff
@@ -80,6 +80,13 @@ Status TitanDBImpl::BackgroundGC(LogBuffer* log_buffer) {
         cfh = db_impl_->GetColumnFamilyHandleUnlocked(column_family_id);
         assert(column_family_id == cfh->GetID());
         blob_gc->SetColumnFamily(cfh.get());
+        if (blob_gc->trigger_next()) {
+          // there is still data remain to be GC
+          // and put this cf back to GC queue and tigger MaybeScheduleGC to wake
+          // up another gc
+          AddToGCQueue(column_family_id);
+          MaybeScheduleGC();
```
Connor1996: seem no need to schedule GC here?

src/blob_gc_picker.cc
GlitterIsMe: because the file is access in descending order so if I meet a file with discard_ratio less than threshold, then the discard ratio of file after this also less than threshold?

src/blob_storage.h
GlitterIsMe: blob_storage.cc uses PRIu64 so blob_storage.h includes <inttypes.h>. blob_gc_picker.cc uses it too and it include blob_storage.h so I think it's not necessary to include in blob_gc_picker.h, is it right? or I need include this file in picker also?

src/blob_gc_picker.cc
Connor1996: on the restart, all the files' score is 1, but the discardable ratio is 0.

src/blob_gc_picker.cc
GlitterIsMe: I think gc_picker stops picking file when the batch_size greater than max_batch_size and output_size is larger than min_output_threshold, to prevent generation of small files, so if && changed to || it would be same as the origin version...(but the strategy maybe too simple, too)

Connor1996: If the output size always small, it would lead to reading too much in a GC round.

src/db_impl_gc.cc
```diff
@@ -14,9 +14,8 @@ void TitanDBImpl::MaybeScheduleGC() {
 
   if (shuting_down_.load(std::memory_order_acquire)) return;
 
-  while (unscheduled_gc_ > 0 &&
+  while (!gc_queue_.empty() &&
```
Connor1996: `gc_queue_` would not pop immediately, this leads to schedule much more than `gc_queue_.size()`

src/blob_gc_picker.cc
```diff
@@ -33,21 +40,48 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
...
+        next_gc_size += blob_file->file_size();
+        if (next_gc_size > cf_options_.min_gc_batch_size) {
+          maybe_continue_next_time = true;
+          ROCKS_LOG_DEBUG(db_options_.info_log,
+                         "remain more than %" PRIu64
+                         " bytes to be gc and trigger after this gc",
+                         next_gc_size);
+          break;
+        }
+      }else{
```
Connor1996: format

src/blob_gc_picker.cc
```diff
@@ -33,21 +40,48 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
     //                   blob_file->marked_for_sample);
 
     if (!CheckBlobFile(blob_file.get())) {
-      ROCKS_LOG_INFO(db_options_.info_log, "file number:%lu no need gc",
+      ROCKS_LOG_DEBUG(db_options_.info_log, "file number:%lu no need gc",
```
Connor1996: this should be info level

src/blob_gc_picker.cc
```diff
@@ -33,21 +40,48 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
...
+        // trigger gc after this
+        stop_picking = true;
+      }
+    } else {
+      if (blob_file->GetDiscardableRatio() >=
+          cf_options_.blob_file_discardable_ratio) {
+        next_gc_size += blob_file->file_size();
+        if (next_gc_size > cf_options_.min_gc_batch_size) {
+          maybe_continue_next_time = true;
+          ROCKS_LOG_DEBUG(db_options_.info_log,
```
Connor1996: this should be info level

src/blob_gc_picker.cc
```diff
@@ -33,21 +40,48 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
...
+                         next_gc_size);
+          break;
+        }
+      }else{
+        break;
+      }
+    }
   }
-
+  ROCKS_LOG_INFO(db_options_.info_log,
```
Connor1996: I think this should be debug level?

src/db_impl_gc.cc
GlitterIsMe: so it's necessary for the unscheduled_gc, and that is why rocksdb define unscheduled_flush and unscheduled_compaction...

src/blob_gc_picker.cc
GlitterIsMe: sorry, forgot it

src/blob_gc_picker.cc
GlitterIsMe: OK~

src/db_impl_gc.cc
Connor1996:
```c++
auto i = std::min((int32_t)gc_queue_.size(), db_options_.max_background_gc - 
 bg_gc_scheduled_);
while (i--) {
```
is not enough?

GlitterIsMe: so...why not using unscheduled_gc_, it is more comprehensible

Connor1996: alright, keep the `unscheduled_gc_`

src/blob_gc.h
```diff
@@ -13,7 +13,7 @@ namespace titandb {
 class BlobGC {
  public:
   BlobGC(std::vector<BlobFileMeta*>&& blob_files,
-         TitanCFOptions&& _titan_cf_options);
+         TitanCFOptions&& _titan_cf_options, bool need_trigger_next = false);
```
yiwu-arbug: seems there's no point to have `need_trigger_next` being optional. I prefer to avoid optional params because the caller may forget to pass them sometimes.

src/db_impl_gc.cc
```diff
@@ -81,9 +81,10 @@ Status TitanDBImpl::BackgroundGC(LogBuffer* log_buffer) {
         cfh = db_impl_->GetColumnFamilyHandleUnlocked(column_family_id);
         assert(column_family_id == cfh->GetID());
         blob_gc->SetColumnFamily(cfh.get());
-        if (blob_gc->trigger_next()) {
+        if (blob_gc->trigger_next() && gc_queue_.size() < 50) {
```
GlitterIsMe: I'm not sure whether to set a value directly or add a field in TitanCFOptions and make it adjustable, and not sure the appropriate value should be here

yiwu-arbug: Instead of limiting the queue size, how about we limit the number of jobs in queue for the CF does not exceed max_background_gc?

src/blob_gc_job.cc
```diff
@@ -149,21 +149,25 @@ Status BlobGCJob::SampleCandidateFiles() {
...
+    Status s = DoSample(file, &selected, estimate_discardable_size);
     if (!s.ok()) {
       return s;
     }
     if (selected) {
       result.push_back(file);
-    }
+    }else{
+      // this update the discardable size of this file according to sample
+      file->AddDiscardableSize(estimate_discardable_size);
```
yiwu-arbug: why? IMO `AddDiscardableSize` is to collect discardable size from table properties of deleted SSTs, which is orthogonal to the result of sampling.

GlitterIsMe: because discardable_size of recovered file is 0 after a re-open so I think maybe I can set an estimated discardable size by the result of sample  if this file is not picked by gc_picker......seems there maybe some mistake, only the file just recovered need this step......or there maybe some other better way(I just think maybe we can use the result of sample)

src/db_impl_gc.cc
GlitterIsMe: but if the number of jobs is too small(like that we set max_bg_gc = 1), trigger_next may never triggered

src/blob_gc_job.cc
yiwu-arbug: Even so, the estimation from sampling is not compatible with discardable size updated from table property. I think we should not mix the two, otherwise the value of discardable size would be hard to reason about.

src/db_impl_gc.cc
yiwu-arbug: max_bg_gc+1 then? Anyway it seems to me the only factor that would affect queue size is how many background GC thread we have.

src/blob_format.h
```diff
@@ -151,7 +155,7 @@ class BlobFileMeta {
   FileState state_{FileState::kInit};
 
   uint64_t discardable_size_{0};
-  //  bool marked_for_gc_{false};
+  bool gc_mark_{false};
```
Connor1996: add some comments

src/blob_format.h
```diff
@@ -137,6 +137,10 @@ class BlobFileMeta {
   bool is_obsolete() const { return state_ == FileState::kObsolete; }
   uint64_t discardable_size() const { return discardable_size_; }
 
+  bool gc_mark() const { return gc_mark_; }
+  void SetGCMark() { gc_mark_ = true; }
```
Connor1996: maybe `set_gc_mark(bool)` ?

src/blob_storage.cc
```diff
@@ -121,8 +121,9 @@ void BlobStorage::ComputeGCScore() {
     gc_score_.push_back({});
     auto& gcs = gc_score_.back();
     gcs.file_number = file.first;
-    if (file.second->file_size() < cf_options_.merge_small_file_threshold) {
-      gcs.score = 1;
+    if (file.second->file_size() < cf_options_.merge_small_file_threshold ||
+        file.second->gc_mark()) {
+      gcs.score = cf_options_.blob_file_discardable_ratio;
```
Connor1996: add some comments

src/blob_gc_picker.cc
```diff
@@ -15,12 +15,25 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
   std::vector<BlobFileMeta*> blob_files;
 
   uint64_t batch_size = 0;
+  uint64_t min_output_threshold =
```
Connor1996: why not use `blob_file_target_size`?

src/blob_gc_picker.cc
Connor1996: @GlitterIsMe

src/blob_format.h
GlitterIsMe: emmm, ok, because I just copy it from your version :P, I will correct it

src/blob_gc_picker.cc
GlitterIsMe: If it stop picking file when the estimate output get blob_file_tartget_size, the input size maybe too large in some scene, so I choose the max value between _small_file_merge_threshold_ and _target_file_size / 4_...

src/blob_format.h
```diff
@@ -151,7 +154,9 @@ class BlobFileMeta {
   FileState state_{FileState::kInit};
 
   uint64_t discardable_size_{0};
-  //  bool marked_for_gc_{false};
+  // gc_mark is set to true when this file is recovered from re-opening the DB
+  // that means this file need to be gc
```
Connor1996:
```suggestion
  // that means this file needs to be checked for GC
```

src/db_impl_gc.cc
```diff
@@ -102,6 +103,15 @@ Status TitanDBImpl::BackgroundGC(LogBuffer* log_buffer) {
       s = blob_gc_job.Finish();
     }
     blob_gc->ReleaseGcFiles();
+
+    if (blob_gc->trigger_next() &&
+        (bg_gc_scheduled_ - 1 + gc_queue_.size() <
+         2 * static_cast<uint32_t>(db_options_.max_background_gc))) {
+      // there is still data remain to be GC
```
Connor1996:
```suggestion
      // there is still data remained to be GCed
```

src/blob_gc.h
```diff
@@ -43,12 +43,15 @@ class BlobGC {
...
 
+  bool trigger_next() { return trigger_next_; }
+
  private:
   std::vector<BlobFileMeta*> inputs_;
   std::vector<BlobFileMeta*> sampled_inputs_;
   std::vector<BlobFileMeta*> outputs_;
   TitanCFOptions titan_cf_options_;
   ColumnFamilyHandle* cfh_{nullptr};
+  const bool trigger_next_;  // whether need to trigger gc after this gc or not
```
Connor1996: Please comment in a new line

src/blob_gc_picker.cc
```diff
@@ -15,12 +15,25 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
...
   //                 blob_storage->gc_score().size());
+  bool stop_picking = false;
+  bool maybe_continue_next_time = false;
+  uint64_t next_gc_size = 0;
   for (auto& gc_score : blob_storage->gc_score()) {
     auto blob_file = blob_storage->FindFile(gc_score.file_number).lock();
     assert(blob_file);
+    if (!blob_file ||
+        blob_file->file_state() == BlobFileMeta::FileState::kBeingGC) {
+      // skip this file id this file is being GC
```
Connor1996: Capitalize the first letter. Please check all the comments.

src/blob_gc_picker.cc
Connor1996: @yiwu-arbug What do you think?

src/db_impl_gc.cc
```diff
@@ -102,6 +103,15 @@ Status TitanDBImpl::BackgroundGC(LogBuffer* log_buffer) {
       s = blob_gc_job.Finish();
     }
     blob_gc->ReleaseGcFiles();
+
+    if (blob_gc->trigger_next() &&
+        (bg_gc_scheduled_ - 1 + gc_queue_.size() <
```
Connor1996: Should consider CF? If the queue is occupied with other CFs and there is no task of current CF, we should still add to GC queue.

src/blob_format.h
```diff
@@ -137,6 +137,9 @@ class BlobFileMeta {
   bool is_obsolete() const { return state_ == FileState::kObsolete; }
   uint64_t discardable_size() const { return discardable_size_; }
 
+  bool gc_mark() const { return gc_mark_; }
+  void SetGCMark(bool mark) { gc_mark_ = mark; }
```
Connor1996: should be `set_gc_mark`

src/db_impl_gc.cc
GlitterIsMe: For my opinion, if we consider the CF, we have to add a variable in DBImpl or ColumnFamily to record bg_gc_scheduled_ of each CF, it makes code complicate. If a gc task is done by previous task, even it triggered nothing would happen, I think it's unnecessary to consider queue size, seems that add to queue directly is OK.

src/blob_gc_picker.cc
```diff
@@ -15,12 +15,25 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
...
+  uint64_t estimate_output_size = 0;
   //  ROCKS_LOG_INFO(db_options_.info_log, "blob file num:%lu gc score:%lu",
   //                 blob_storage->NumBlobFiles(),
   //                 blob_storage->gc_score().size());
+  bool stop_picking = false;
+  bool maybe_continue_next_time = false;
+  uint64_t next_gc_size = 0;
   for (auto& gc_score : blob_storage->gc_score()) {
     auto blob_file = blob_storage->FindFile(gc_score.file_number).lock();
     assert(blob_file);
```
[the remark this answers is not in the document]
GlitterIsMe: here?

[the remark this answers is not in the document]
Connor1996: The assert is not always true?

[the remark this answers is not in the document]
GlitterIsMe: ok, got it

src/blob_gc_job_test.cc
```diff
@@ -185,7 +185,7 @@ class BlobGCJobTest : public testing::Test {
     auto rewrite_status = base_db_->Write(WriteOptions(), &wb);
 
     std::vector<BlobFileMeta*> tmp;
-    BlobGC blob_gc(std::move(tmp), TitanCFOptions());
+    BlobGC blob_gc(std::move(tmp), TitanCFOptions(), false);
```
yiwu-arbug: nit: add comment like:
`BlobGC blob_gc(..., false/*trigger_next*/)`

src/blob_storage.h
```diff
@@ -65,6 +66,7 @@ class BlobStorage {
   void MarkAllFilesForGC() {
     MutexLock l(&mutex_);
     for (auto& file : files_) {
+      file.second->set_gc_mark(true);
```
yiwu-arbug: hmm, sorry for notice this late, but since we have the `kDbRestart` state, why we need an extra boolean flag `gc_mark`?

src/db_impl.h
```diff
@@ -149,6 +149,7 @@ class TitanDBImpl : public TitanDB {
...
```
yiwu-arbug: add `mutex_.AssertHeld()`?

src/blob_storage.h
GlitterIsMe: because FileState will set to normal when FileStateTraist received FileEvent::kDbRestart, we cannot know a normal file is generated by restart or compaction

src/db_impl.h
GlitterIsMe: OK

src/blob_storage.h
yiwu-arbug: Got it. I misunderstand the code.

PR #37, opened 2019-07-10
Check edits before written to manifest

Connor1996:
Now we check edit validity after written to manifest. That is to say, if the validity check does not pass for some edits, then the restart of Titan would always fail because the invalid edits are already written to manifest.

### DIFF
--- a/src/edit_collector.h
+++ b/src/edit_collector.h
@@ -17,22 +17,22 @@ class EditCollector {
     auto cf_id = edit.column_family_id_;
     auto& collector = column_families_[cf_id];
 
-    Status s;
     for (auto& file : edit.added_files_) {
-      s = collector.AddFile(file);
-      if (!s.ok()) return s;
+      status_ = collector.AddFile(file);
+      if (!status_.ok()) return status_;
     }
     for (auto& file : edit.deleted_files_) {
-      collector.DeleteFile(file.first, file.second);
-      if (!s.ok()) return s;
+      status_ = collector.DeleteFile(file.first, file.second);
+      if (!status_.ok()) return status_;
     }
 
     if (edit.has_next_file_number_) {
       if (edit.next_file_number_ < next_file_number_) {
-        return Status::Corruption("Edit has a smaller next file number " +
-                                  ToString(edit.next_file_number_) +
-                                  " than current " +
-                                  ToString(next_file_number_));
+        status_ =
+            Status::Corruption("Edit has a smaller next file number " +
+                               ToString(edit.next_file_number_) +
+                               " than current " + ToString(next_file_number_));
+        return status_;
       }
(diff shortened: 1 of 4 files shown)


Connor1996:
PTAL @yiwu-arbug @GlitterIsMe

src/edit_collector.h
```diff
@@ -17,30 +17,64 @@ class EditCollector {
...
+    return Status::OK();
+  }
+
   Status Apply(VersionSet& vset) {
+    if (!status_.ok()) {
+      return status_;
+    }
+
+    if (!checked_) {
+      status_ = Check(vset);
```
yiwu-arbug: The API is a bit confusing since caller can call `AddEdit` after `Check`, where we should double check for error. I think we can rename `Check` to `Seal` and return error if `AddEdit` is called after sealed.

src/edit_collector.h
```diff
@@ -126,6 +164,30 @@ class EditCollector {
...
+
+      for (auto& file : deleted_files_) {
+        auto number = file.first;
+        // just skip paired added and deleted files
+        if (added_files_.count(number) > 0) {
+          continue;
+        }
+        auto blob = storage->FindFile(number).lock();
+        if (!blob) {
+          return Status::NotFound("invalid file number");
```
yiwu-arbug: add the file number in the error message?

src/edit_collector.h
```diff
@@ -88,7 +124,7 @@ class EditCollector {
       return Status::OK();
     }
 
-    Status Apply(shared_ptr<BlobStorage>& storage) {
+    Status Check(shared_ptr<BlobStorage>& storage) {
```
yiwu-arbug: Pass raw pointer of `storage` instead? Since we don't care about ownership.

src/edit_collector.h
```diff
@@ -126,6 +164,30 @@ class EditCollector {
           return Status::Corruption("Blob file " + ToString(number) +
                                     " has been deleted already");
         }
+      }
+
+      return Status::OK();
+    }
+
+    Status Apply(shared_ptr<BlobStorage>& storage) {
```
yiwu-arbug: ditto

src/edit_collector.h
```diff
@@ -11,9 +11,17 @@
...
 
+// A collector to apply edits in batch.
+// The functions should be called in the sequence:
+//    AddEdit() -> Seal() -> Apply()
 class EditCollector {
  public:
+  // Add the edit into the batch.
   Status AddEdit(const VersionEdit& edit) {
+    if (sealed_)
+      return Status::Corruption(
```
yiwu-arbug: `Status::Incomplete` might be a little bit more appropriate.

Connor1996:
CI failure is irrelevant, so merge

PR #39, opened 2019-07-10
Replace atomic var by normal var

GlitterIsMe:
Because db is locked when  add/sub  bg_gc_scheduled_, so this variable is no need to be atomic

### DIFF
--- a/src/db_impl.h
+++ b/src/db_impl.h
@@ -200,7 +200,7 @@ class TitanDBImpl : public TitanDB {
   // gc_queue_ hold column families that we need to gc.
   // pending_gc_ hold column families that already on gc_queue_.
   std::deque<uint32_t> gc_queue_;
-  
+
   // Guarded by mutex_.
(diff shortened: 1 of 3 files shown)


disksing:
Thanks for your contribution @GlitterIsMe 
/cc @yiwu-arbug @Connor1996

src/db_impl.h
```diff
@@ -200,8 +200,9 @@ class TitanDBImpl : public TitanDB {
   // gc_queue_ hold column families that we need to gc.
   // pending_gc_ hold column families that already on gc_queue_.
   std::deque<uint32_t> gc_queue_;
-
-  std::atomic_int bg_gc_scheduled_{0};
+  
+  // Guarded by mutex_.
+  int bg_gc_scheduled_{0};
 
   std::atomic_bool shuting_down_{false};
```
Connor1996: how about this?

PR #44, opened 2019-07-15
[DNM] DeletefilesInRanges + trigger next

Connor1996:
only for test


src/blob_storage.cc
```diff
@@ -121,8 +121,9 @@ void BlobStorage::ComputeGCScore() {
     gc_score_.push_back({});
     auto& gcs = gc_score_.back();
     gcs.file_number = file.first;
-    if (file.second->file_size() < cf_options_.merge_small_file_threshold) {
-      gcs.score = 1;
+    if (file.second->file_size() <= cf_options_.merge_small_file_threshold ||
+        file.second->gc_mark()) {
+      gcs.score = cf_options_.blob_file_discardable_ratio;
```
yiwu-arbug: why not set to 1 as before?

src/blob_format.h
```diff
@@ -151,7 +155,7 @@ class BlobFileMeta {
   FileState state_{FileState::kInit};
 
   uint64_t discardable_size_{0};
-  //  bool marked_for_gc_{false};
+  bool gc_mark_{false};
```
yiwu-arbug: Add a comment describe meaning of `gc_mark_`

src/blob_gc_picker.cc
```diff
@@ -57,8 +57,10 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
         stop_picking = true;
       }
     } else {
-      if (blob_file->GetDiscardableRatio() >=
-          cf_options_.blob_file_discardable_ratio) {
+      if (blob_file->file_size() <= cf_options_.merge_small_file_threshold ||
```
yiwu-arbug: Does it mean we are more aggressively merging small files than before?

src/blob_storage.cc
Connor1996: To make them low priority. We want to merge blob files that exceed the discardable ratio first.

src/blob_gc_picker.cc
Connor1996: Yep, merging small file doesn't work as expected before cause `DoSample` doesn't consider `merge_small_file_threshold`.

yiwu-arbug: I see.

src/blob_storage.cc
yiwu-arbug: make sense. better to add an inline comment.

PR #46, opened 2019-07-22
Change raw ptr in BlobGC to shared_ptr

GlitterIsMe:
**Problem**: I find that titan_db_test failed at the BackgroundErrorTrigger case with ASAN ON under the trigger_next pr 
**Cause**: After debugging I find that in the BlobGCJob::InstallOutputFiles, BlobFileMeta was generated as shared_ptr, but add to BlobGCJoB::outputs_ as raw_ptr( by shared_ptr::get()). When writing to manifest failed in the BatchFinishFiles, shared_ptr of BlobFileMeta may be released and the access of raw pointer hold by outputs_ may cause error(use memory had been free)
**Fix**: I make the input_, output_ and sampled_input in BlobGCJon from raw_ptr to shared_ptr and modified  related interface
**Note**: This problem was find from another pr(trigger next), it maybe caused by adding small file to BlobGC directly, and master branch can pass the titan_db_test, so I am not sure whether this fix is necessary(even so I think we'd better not take raw_ptr from shared_ptr)

### DIFF
--- a/src/blob_gc_job.cc
+++ b/src/blob_gc_job.cc
@@ -475,7 +475,7 @@ Status BlobGCJob::InstallOutputBlobFiles() {
           builder.first->GetNumber(), builder.first->GetFile()->GetFileSize());
 
       if (!tmp.empty()) {
-        tmp.append("");
+        tmp.append(" ");
       }
       tmp.append(std::to_string(file->file_number()));
       // To temporarily hold the ptr of BlobFileMeta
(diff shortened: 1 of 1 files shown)


yiwu-arbug:
+1 for checking return status of `BatchFinishFiles`. And we can move `blob_gc_->AddOutputFile` to after `BatchFinishFiles` success, when we know the file meta has a owner. shared_ptr is not super performant and should not be used blindly.

src/blob_gc_job.cc
```diff
@@ -477,15 +478,22 @@ Status BlobGCJob::InstallOutputBlobFiles() {
...
       files.emplace_back(std::make_pair(file, std::move(builder.first)));
     }
     ROCKS_LOG_BUFFER(log_buffer_, "[%s] output[%s]",
                      blob_gc_->column_family_handle()->GetName().c_str(),
                      tmp.c_str());
-    this->blob_file_manager_->BatchFinishFiles(
+    s = this->blob_file_manager_->BatchFinishFiles(
         blob_gc_->column_family_handle()->GetID(), files);
+    if (s.ok()) {
+      for (auto file : tmp_file_metas) {
```
Connor1996: why not use file meta in `files`?

GlitterIsMe: this will cause compile error...I have not figured out the reason

GlitterIsMe: I figured out it, because I use `for(auto file : files)`, but files is consisted of pair<share_ptr, unique_ptr> in which unique_ptr cannot be copied, correct way is `for(auto &file : files)`

PR #51, opened 2019-07-30
Add delete blob files in ranges

Connor1996:
- Add smallest and largest key in `BlobFileMeta`
- Use map to record ranges of blob files
- Delete blobs in ranges in `DeleteFilesInRanges` API


Connor1996:
PTAL @yiwu-arbug  @GlitterIsMe

src/blob_storage.cc
```diff
@@ -24,6 +24,26 @@ Status BlobStorage::NewPrefetcher(uint64_t file_number,
                                     result);
 }
 
+Status BlobStorage::DeleteBlobsInRanges(const RangePtr* ranges, size_t n,
+                                        bool include_end,
+                                        SequenceNumber obsolete_sequence) {
+  MutexLock l(&mutex_);
+  for (size_t i = 0; i < n; i++) {
+    auto begin = ranges[i].start, end = ranges[i].limit;
```
yiwu-arbug: One variable per line? 

Also change `auto` to `Slice*`? For simpler types I prefer writing out the type instead of `auto` for readability.

src/blob_storage.cc
```diff
@@ -24,6 +24,26 @@ Status BlobStorage::NewPrefetcher(uint64_t file_number,
...
+    auto cmp = cf_options_.comparator;
+
+    for (auto it = blob_ranges_.lower_bound(*begin);
+         it != blob_ranges_.upper_bound(*end); it++) {
+      if (it->second->is_obsolete()) continue;
+      if ((include_end && cmp->Compare(it->second->largest_key(), *end) <= 0) ||
```
yiwu-arbug: don't we need to check if `largest_key()` is empty?

yiwu-arbug: assert `smallest_key()` is empty or `>= begin`

src/blob_storage.h
```diff
@@ -115,6 +129,23 @@ class BlobStorage {
 
   // Only BlobStorage OWNS BlobFileMeta
   std::unordered_map<uint64_t, std::shared_ptr<BlobFileMeta>> files_;
+
+  class InternalComparator {
+   public:
+    InternalComparator() = default;
+    InternalComparator(const Comparator* comparator)
```
yiwu-arbug: remove the default constructor and add `explicit` keyword for the constructor with one param.

src/blob_storage.cc
```diff
@@ -80,33 +136,22 @@ void BlobStorage::GetObsoleteFiles(std::vector<std::string>* obsolete_files,
...
       // remove obsolete files
-      auto p = files_.find(file_number);
-      assert(p != files_.end());
-      file_dropped++;
-      file_dropped_size += p->second->file_size();
-      files_.erase(p);
-
-      file_cache_->Evict(file_number);
-
+      auto b = RemoveFile(file_number);
```
yiwu-arbug:
```
bool __attribute__((__unused__)) removed = RemoveFiles(file_number);
assert(removed);
```
`__attribute__((__unused__))` is to avoid unused variable warning in release build.

src/blob_storage.cc
```diff
@@ -66,6 +99,29 @@ void BlobStorage::MarkFileObsolete(std::shared_ptr<BlobFileMeta> file,
...
+  // Removes from blob_ranges_
+  auto p = blob_ranges_.equal_range(file->second->smallest_key());
+  for (auto it = p.first; it != p.second; it++) {
+    if (it->second->file_number() == file->second->file_number()) {
+      it = blob_ranges_.erase(it);
+      break;
+    }
+  }
+  files_.erase(file_number);
+  file_cache_->Evict(file_number);
```
yiwu-arbug: Shall evict from cache before doing anything. Or does it matter?

src/blob_storage.h
```diff
@@ -7,13 +7,14 @@
 #include "blob_format.h"
 #include "blob_gc.h"
 #include "rocksdb/options.h"
+#include "table/internal_iterator.h"
```
yiwu-arbug: Why this is needed?

src/blob_storage.h
```diff
@@ -60,6 +66,11 @@ class BlobStorage {
     return files_.size();
   }
 
+  std::size_t NumObsoleteBlobFiles() const {
```
yiwu-arbug: Help me add a `TODO` to return directly from this function for the obsolete files DB property?

src/blob_file_iterator.h
```diff
@@ -93,19 +93,25 @@ class BlobFileMergeIterator {
   BlobIndex GetBlobIndex() { return current_->GetBlobIndex(); }
 
  private:
-  class IternalComparator {
+  class BlobFileIterComparator {
    public:
+    BlobFileIterComparator() = default;
```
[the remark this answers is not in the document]
Connor1996: No, `std::priority_queue()` needs calling `BlobFileIterComparator()`

src/blob_storage.cc
Connor1996: good catch!

Connor1996: why assert `smallest_key()` is empty? it should be included in `>= begin`.

src/blob_format.cc
```diff
@@ -132,11 +132,27 @@ bool operator==(const BlobIndex& lhs, const BlobIndex& rhs) {
...
+  PutLengthPrefixedSlice(dst, largest_key_);
 }
 
 Status BlobFileMeta::DecodeFrom(Slice* src) {
   if (!GetVarint64(src, &file_number_) || !GetVarint64(src, &file_size_)) {
-    return Status::Corruption("BlobFileMeta Decode failed");
+    return Status::Corruption("BlobFileMeta decode failed");
+  }
+  // old version may not have these fields.
+  if (src->size() > 0) {
```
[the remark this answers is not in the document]
Connor1996: oh, my mistake. Seems should add some related test.

src/blob_storage.cc
Connor1996: any difference? what's the matter?

src/blob_storage.cc
Connor1996: and `>= begin` is a promise of `lower_bound`, is it really necessary?

src/blob_file_iterator.h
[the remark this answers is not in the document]
yiwu-arbug: ok

src/blob_storage.cc
yiwu-arbug: Sorry I didn't explain well. Can a reader accidentally obtain file handle from `file_cache_` after we remove the file from `file_` and if so will it cause error?

src/blob_storage.cc
yiwu-arbug: My understanding is, `smallest_key()` is empty means it's of legacy manifest and so assume to be cover the full key range (more like `smallest_key() < begin` for any `begin`). Am I missing something?

yiwu-arbug: Just an assert to verify our data-structure is correct in test or debug mode.

src/blob_storage.cc
```diff
@@ -24,6 +24,38 @@ Status BlobStorage::NewPrefetcher(uint64_t file_number,
...
+    // nullptr means the minimum or maximum.
+    for (auto it = ((begin) ? blob_ranges_.lower_bound(*begin)
+                            : blob_ranges_.begin());
+         it != ((end) ? blob_ranges_.upper_bound(*end) : blob_ranges_.end());
+         it++) {
+      // Obsolete files are to be deleted, so just skip.
+      if (it->second->is_obsolete()) continue;
+      // The smallest and largest key of blob file meta of the old version are
+      // empty, so skip.
+      if (it->second->largest_key().empty() && end) continue;
```
yiwu-arbug: ditto, `end != nullptr`.

src/blob_storage.cc
```diff
@@ -45,14 +77,27 @@ void BlobStorage::ExportBlobFiles(
 void BlobStorage::AddBlobFile(std::shared_ptr<BlobFileMeta>& file) {
   MutexLock l(&mutex_);
   files_.emplace(std::make_pair(file->file_number(), file));
+  blob_ranges_.emplace(std::make_pair(file->smallest_key(), file));
```
yiwu-arbug: exclude files with empty smallest key?

src/blob_format.cc
```diff
@@ -132,11 +132,33 @@ bool operator==(const BlobIndex& lhs, const BlobIndex& rhs) {
...
+  return Status::OK();
 }
 
 Status BlobFileMeta::DecodeFrom(Slice* src) {
   if (!GetVarint64(src, &file_number_) || !GetVarint64(src, &file_size_)) {
-    return Status::Corruption("BlobFileMeta Decode failed");
+    return Status::Corruption("BlobFileMeta decode failed");
+  }
+  Slice str;
+  if (GetLengthPrefixedSlice(src, &str)) {
```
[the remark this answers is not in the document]
Connor1996: `str.size() == 0` iff `GetLengthPrefixedSlice() == false`. Is it necessary to add it? maybe just add an assert?

src/blob_storage.cc
Connor1996: Nope. When DeleteBlobFilesInRange with range(nullptr, nullptr), we should delete these files as well.

src/blob_storage.cc
Connor1996: `Evict` actually calls `cache_->Erase` which tells the underlying entry will be kept around until all existing handles to it have been released. So it would cause an error. 
And seems the case `a reader accidentally obtain file handle from file_cache_ after we remove the file from file_ ` would happen, cause we remove file only if all snapshot exceed the obsolete sequence.

src/blob_storage.cc
yiwu-arbug: got it.

src/blob_format.cc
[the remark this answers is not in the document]
yiwu-arbug: `str.size` can be 0 if the encoded length is 0, where `GetLengthPrefixedSlice` will return true.

src/blob_format.h
```diff
@@ -9,17 +9,38 @@
...
+// ...
+// [blob head + record N]
+// [meta block 1]
+// [meta block 2]
+// ...
+// [meta block K]
+// [meta index block]
+// [blob file footer]
+
+// Format of blob head (9 bytes):
```
yiwu-arbug: Thanks for adding the illustration!

src/blob_format.cc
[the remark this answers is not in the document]
Connor1996: seems it not right? if the encoded length is 0, `GetLengthPrefixedSlice` should return false.

yiwu-arbug:
@Connor1996 I remember you mention there's an additional issue with the patch. What is it and how is it/will it be fixed?

Connor1996:
@yiwu-arbug The files deleted by DeleteBlobFilesInRange are not recorded in manifest, I will fix it and add a test in this patch later.

src/blob_file_set.cc
```diff
@@ -284,6 +284,34 @@ Status BlobFileSet::MaybeDestroyColumnFamily(uint32_t cf_id) {
...
+    Status s = it->GetBlobFilesInRanges(ranges, n, include_end, &files);
+    if (!s.ok()) return s;
+    for (auto file_number : files) {
+      edit->DeleteBlobFile(file_number);
+    }
+    s = LogAndApply(edit);
+    if (!s.ok()) return s;
+
+    ROCKS_LOG_INFO(db_options_.info_log,
+                   "%d blob files are deleted in the range [%s, %s%c",
```
yiwu-arbug: print all the files being deleted?

Connor1996: try to test `auto-merge` bot, will add some logs in other PR later

PR #52, opened 2019-07-31
init metrics_ in BlobGCJob

GlitterIsMe:
fix: init fields in metrics, these metrics would be random value if not init it


yiwu-arbug:
nice

PR #54, opened 2019-08-01
Limit blob file read IO with rate_limiter

wujy-cs:
Limit Titan blob file read IO with rate_limiter, to avoid burst of IO starve online workloads.

### DIFF
--- a/src/blob_file_cache.cc
+++ b/src/blob_file_cache.cc
@@ -73,9 +73,7 @@ Status BlobFileCache::FindFile(uint64_t file_number, uint64_t file_size,
     if (db_options_.advise_random_on_open) {
       f->Hint(RandomAccessFile::RANDOM);
     }
-    file.reset(new RandomAccessFileReader(std::move(f), file_name, nullptr,
-                                          nullptr, 0, nullptr,
-                                          env_options_.rate_limiter, true));
+    file.reset(new RandomAccessFileReader(std::move(f), file_name));
   }
 
   std::unique_ptr<BlobFileReader> reader;
--- a/src/blob_file_cache.cc
+++ b/src/blob_file_cache.cc
@@ -73,7 +73,9 @@ Status BlobFileCache::FindFile(uint64_t file_number, uint64_t file_size,
     if (db_options_.advise_random_on_open) {
       f->Hint(RandomAccessFile::RANDOM);
     }
-    file.reset(new RandomAccessFileReader(std::move(f), file_name));
+    file.reset(new RandomAccessFileReader(std::move(f), file_name, nullptr,
+                                          nullptr, 0, nullptr,
+                                          env_options_.rate_limiter, true));
   }
 
   std::unique_ptr<BlobFileReader> reader;
--- a/src/blob_file_test.cc
+++ b/src/blob_file_test.cc
@@ -88,10 +88,6 @@ class BlobFileTest : public testing::Test {
     const int n = 100;
     std::vector<BlobHandle> handles(n);
 
-    auto rate_limiter =
-        std::shared_ptr<RateLimiter>(NewGenericRateLimiter(1 << 30));
-    env_options_.rate_limiter = rate_limiter.get();
-
     std::unique_ptr<WritableFileWriter> file;
     {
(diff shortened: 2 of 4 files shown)


wujy-cs:
@yiwu-arbug Maybe we should use another limiter for reader?

src/blob_file_test.cc
```diff
@@ -88,6 +88,10 @@ class BlobFileTest : public testing::Test {
     const int n = 100;
     std::vector<BlobHandle> handles(n);
 
+    auto rate_limiter =
```
yiwu-arbug: For test, add a new test case? The test should create a mock test rate limiter, ran GC job, and check if it call into the mock rate limiter. It should test with all three values of `RateLimiter::Mode`.

src/blob_file_cache.cc
```diff
@@ -73,7 +73,9 @@ Status BlobFileCache::FindFile(uint64_t file_number, uint64_t file_size,
     if (db_options_.advise_random_on_open) {
       f->Hint(RandomAccessFile::RANDOM);
     }
-    file.reset(new RandomAccessFileReader(std::move(f), file_name));
+    file.reset(new RandomAccessFileReader(std::move(f), file_name, nullptr,
+                                          nullptr, 0, nullptr,
+                                          env_options_.rate_limiter, true));
```
yiwu-arbug: To be consistent with rocksdb, we should rate limit GC only, not user reads. Seems `BlobGCJob` currently call `NewBlobFileReader` directly, which does not go through `BlobFileCache`. Let's revert the change here.

yiwu-arbug:
> @yiwu-arbug Maybe we should use another limiter for reader?

Not really. Check `RateLimiter::Mode`.

wujy-cs:
@thanks!

wujy-cs:
@yiwu-arbug Thanks! Please take a look again?

src/blob_gc_job_test.cc
```diff
@@ -260,6 +344,8 @@ TEST_F(BlobGCJobTest, DiscardEntry) { TestDiscardEntry(); }
 
 TEST_F(BlobGCJobTest, RunGC) { TestRunGC(); }
 
+TEST_F(BlobGCJobTest, GCLimiter) { TestGCLimiter(); }
```
yiwu-arbug: I think we don't need to have a separate `TestGCLimiter()`. Just put the test code inside `BlobGCJobTest` body.

src/blob_gc_job_test.cc
```diff
@@ -196,6 +210,76 @@ class BlobGCJobTest : public testing::Test {
     DestroyDB();
   }
 
+  void TestGCLimiter() {
+    class TestLimiter : public RateLimiter {
+     public:
+      TestLimiter(RateLimiter::Mode mode)
+          : RateLimiter(mode), read(false), write(false) {}
+
+      virtual size_t RequestToken(size_t bytes, size_t alignment,
```
yiwu-arbug: remove `virtual` keyword and add `override` keyword. Same for other methods.

src/blob_gc_job_test.cc
```diff
@@ -196,6 +210,76 @@ class BlobGCJobTest : public testing::Test {
...
+      }
+
+      virtual int64_t GetBytesPerSecond() const { return 0; }
+
+      bool ReadRequested() { return read; }
+
+      bool WriteRequested() { return write; }
+
+     private:
+      bool read;
```
yiwu-arbug: need to reset both of these flags to false after flush, since flush also make use of rate limiter.

src/blob_gc_job_test.cc
```diff
@@ -260,6 +261,98 @@ TEST_F(BlobGCJobTest, DiscardEntry) { TestDiscardEntry(); }
...
+    bool ReadRequested() { return read; }
+
+    bool WriteRequested() { return write; }
+
+   private:
+    bool read;
+    bool write;
+  };
+
+  auto TriggerGC = [this] {
```
yiwu-arbug: give it a better name? It's not triggering GC now.

PR #56, opened 2019-08-11
Dig hole

lhy1024:
add more unit tests


yiwu-arbug:
Forgive my laziness for only reading the proposal but not the source code. I find the approach interesting, and would like to know your thinking around some details:
1. With the approach how shall we handle snapshots? In current implementation, by the time `DiscardEntry()` report a key being discardable, it may be still visible to an old snapshot.
2. AFAIK punch hole is 4k-aligned (e.g. if you punch a hole with size < 4k, you may end up punching nothing). How shall we take it into consideration?
3. How do we estimate WA with the approach?

lhy1024:
@yiwu-arbug Thanks.
In fact, it is a project of the talent plan, and it is not perfect because of the lack of time.
1. There is really no way to deal with snapshots with punching holes.
2. This problem was later discovered by us, using a 4k alignment approach.
3. This has been measured in the updated proposal, theoretically optimizing 10% of the write amplification, but because of the 4k alignment, it is even larger in some cases.

PR #57, opened 2019-08-12
Add benchmark script

YangKeao:
Add script for run benchmark. It a modified version of [benchmark.sh](https://github.com/pingcap/rocksdb/blob/tikv-3.0/tools/benchmark.sh).

- Replace `db_bench` with `titandb_bench`
- Add argument `use_titan` according to environment variable `TITAN`
- Replace implementation of memtable in seqfill and bulkload from vector into skiplist. This is because GC in titan needs GET but vector can lead to poor GET performance.


Connor1996:
PTAL @yiwu-arbug

yiwu-arbug:
LGTM, although I have never use this script before..

PR #59, opened 2019-08-13
[DNM] Add punching hole method for gc

lhy1024:


doc/2019-08-04-write-Amplification.md
```diff
@@ -0,0 +1,208 @@
...
+</p>
+
+<p align="center">
+  <img src="./imgs/1565840407310.png"><br> 
+  Figure 4: OPS
+</p>
+
+#### Analysis
+
+We can see that in both cases of 2k and 4k, the use of hole punching instead introduces a larger write amplification. Because we use 4k alignment to process data, which leads to the storage of one such record need more blocks to be filled zero. And the data we write is not exactly 2k or 4k, considering the head and key. In fact, it is just a little bigger than 2k or 4k.
```
yiwu-arbug: Just to put a note here. To avoid the need to align records, we can have a key->offset map at the end of each blob file, and have GC job iterate through this map instead. The map can probably also reduce GC read amp.

lhy1024: Yes, we also had have such a plan, but at the time we only had a few days left for talent plan, so we chose a method that was easier to implement.
However, misalignment may result in a lot less space for punching holes, because cross-block records will be more.

PR #61, opened 2019-08-22
Upgrade RocksDB to 6.4

yiwu-arbug:
Summary:
Update the code to base off RocksDB 6.4.x. There are a few internal interface changes and files moving around from 5.18 to 6.4. titandb_bench is also updated to base off the version in 6.4.x.

Test Plan:
Travis

### DIFF
--- a/src/blob_file_iterator.cc
+++ b/src/blob_file_iterator.cc
@@ -8,18 +8,20 @@ namespace titandb {
 
 BlobFileIterator::BlobFileIterator(
     std::unique_ptr<RandomAccessFileReader>&& file, uint64_t file_name,
-    uint64_t file_size, const TitanCFOptions& titan_cf_options)
+    uint64_t file_size, const TitanCFOptions& titan_cf_options, bool for_gc)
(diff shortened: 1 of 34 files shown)


src/blob_file_iterator.cc
```diff
@@ -29,8 +31,9 @@ bool BlobFileIterator::Init() {
     return false;
   }
   char footer_buf[BlobFileFooter::kEncodedLength];
-  status_ = file_->Read(file_size_ - BlobFileFooter::kEncodedLength,
-                        BlobFileFooter::kEncodedLength, &slice, footer_buf);
+  status_ =
+      file_->Read(file_size_ - BlobFileFooter::kEncodedLength,
```
Connor1996: seems `for_gc_` is a hint for `RandomFileAccessReader` to decide whether to perform flow control?
Can we always pass `true` here cause `BlobFileIterator` is always used for GC?

yiwu-arbug: Didn't realize `BlobFileIterator` is only for GC. Updated.

PR #66, opened 2019-09-04
Implement basic level merge for blob files

wujy-cs:
- Record blob file's level, which is decided by target level of the compaction or flush that generates this blob file.
- Add a option to enable level merge. With level merge enabled, titan table builder will merge values of blob files to a new blob file if:
    - blob file's level lower than builder target level, and
    - target level is higher than or equals to merge level. we set merge level to max(1, cf's highest level-1) for now.
- Limit blob file size to blob_file_target_size in table builder, to avoid generate large blob files during level merge.

### DIFF
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -118,6 +118,12 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // Default: kNormal
   TitanBlobRunMode blob_run_mode{TitanBlobRunMode::kNormal};
 
+  // If set true, values will be merged to a new blob file while their
+  // corresponding keys are compacted to last two level in LSM-Tree.
+  //
+  // Default: false
+  bool level_merge{false};
+
   TitanCFOptions() = default;
   explicit TitanCFOptions(const ColumnFamilyOptions& options)
       : ColumnFamilyOptions(options) {}
--- a/src/blob_format.cc
+++ b/src/blob_format.cc
@@ -133,10 +133,13 @@ bool operator==(const BlobIndex& lhs, const BlobIndex& rhs) {
 void BlobFileMeta::EncodeTo(std::string* dst) const {
   PutVarint64(dst, file_number_);
   PutVarint64(dst, file_size_);
+  PutVarint64(dst, file_entries_);
+  PutVarint32(dst, file_level_);
 }
 
 Status BlobFileMeta::DecodeFrom(Slice* src) {
-  if (!GetVarint64(src, &file_number_) || !GetVarint64(src, &file_size_)) {
+  if (!GetVarint64(src, &file_number_) || !GetVarint64(src, &file_size_) ||
+      !GetVarint64(src, &file_entries_) || !GetVarint32(src, &file_level_)) {
     return Status::Corruption("BlobFileMeta Decode failed");
   }
   return Status::OK();
@@ -144,7 +147,9 @@ Status BlobFileMeta::DecodeFrom(Slice* src) {
 
 bool operator==(const BlobFileMeta& lhs, const BlobFileMeta& rhs) {
   return (lhs.file_number_ == rhs.file_number_ &&
-          lhs.file_size_ == rhs.file_size_);
+          lhs.file_size_ == rhs.file_size_ &&
+          lhs.file_entries_ == rhs.file_entries_ &&
+          lhs.file_level_ == rhs.file_level_);
 }
 
 void BlobFileMeta::FileStateTransit(const FileEvent& event) {
--- a/src/table_builder.cc
+++ b/src/table_builder.cc
@@ -60,6 +60,37 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
       AppendInternalKey(&index_key, ikey);
       base_builder_->Add(index_key, index_value);
     }
+  } else if (ikey.type == kTypeBlobIndex && cf_options_.level_merge &&
+             target_level_ >= merge_level_) {
+    // we merge value to new blob file
+    BlobIndex index;
+    Slice copy = value;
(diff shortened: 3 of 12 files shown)


src/blob_format.cc
```diff
@@ -133,18 +133,23 @@ bool operator==(const BlobIndex& lhs, const BlobIndex& rhs) {
...
   PutVarint64(dst, file_number_);
   PutVarint64(dst, file_size_);
+  PutVarint64(dst, file_entries_);
+  PutVarint32(dst, file_level_);
 }
 
 Status BlobFileMeta::DecodeFrom(Slice* src) {
-  if (!GetVarint64(src, &file_number_) || !GetVarint64(src, &file_size_)) {
+  if (!GetVarint64(src, &file_number_) || !GetVarint64(src, &file_size_) ||
+      !GetVarint64(src, &file_entries_) || !GetVarint32(src, &file_level_)) {
```
Connor1996: you should consider the compatibility issue

src/table_builder.h
```diff
@@ -46,8 +49,17 @@ class TitanTableBuilder : public TableBuilder {
...
+  bool ShouldMerge(const std::shared_ptr<BlobFileMeta>& file);
+
+  void FinishBlob();
+
+  friend class TableBuilderTest;
+
   Status status_;
   uint32_t cf_id_;
+  std::unique_ptr<ColumnFamilyHandle> cf_handle_;
+  int merge_level_;
```
Connor1996: Please add comments

src/table_builder.h
```diff
@@ -46,8 +49,17 @@ class TitanTableBuilder : public TableBuilder {
...
+  int target_level_;
```
Connor1996: ditto

src/table_builder.cc
```diff
@@ -60,6 +60,37 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+      if (s.ok()) {
+        std::string index_value;
+        AddBlob(ikey.user_key, record.value, &index_value);
+        if (ok()) {
+          std::string index_key;
+          ikey.type = kTypeBlobIndex;
+          AppendInternalKey(&index_key, ikey);
+          base_builder_->Add(index_key, index_value);
+        }
+      } else {
```
Connor1996: verbose `base_builder_->Add(key, value)`

src/blob_format.cc
yiwu-arbug: We should add new metadata to `kAddedBlobFileV2`. See https://github.com/pingcap/titan/pull/51. Let's reuse the same  tag `kAddedBlobFileV2` for both of the changes.

src/table_builder.cc
yiwu-arbug: agree. let's refactor the logic here a bit to avoid multiple `base_builder_->Add(key, value)` calls.

include/titan/options.h
```diff
@@ -124,6 +124,12 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // Default: kNormal
   TitanBlobRunMode blob_run_mode{TitanBlobRunMode::kNormal};
 
+  // If set true, values will be merged to a new blob file while their
+  // corresponding keys are compacted to last two level in LSM-Tree.
```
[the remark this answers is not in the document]
wujy-cs: I'm considering it, or you can give it a new name. :)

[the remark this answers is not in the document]
yiwu-arbug: Not a bad name :)

wujy-cs:
@yiwu-arbug Please take another look

src/table_builder_test.cc
```diff
@@ -318,6 +344,121 @@ TEST_F(TableBuilderTest, NumEntries) {
...
+  const int n = 255;
+  for (unsigned char i = 0; i < n; i++) {
+    std::string key(1, i);
+    InternalKey ikey(key, 1, kTypeValue);
+    std::string value(kMinBlobSize, i);
+    table_builder->Add(ikey.Encode(), value);
+  }
+  ASSERT_OK(table_builder->Finish());
+  uint64_t file_size = 0;
+  ASSERT_OK(env_->GetFileSize(blob_name_, &file_size));
```
yiwu-arbug: I look at the code and seems the test table builder don't write to file with name as `blob_name_`. Am I missing something?

src/table_builder_test.cc
```diff
@@ -318,6 +344,121 @@ TEST_F(TableBuilderTest, NumEntries) {
...
+  ASSERT_OK(table_builder->Finish());
+  ASSERT_OK(base_file->Sync(true));
+  ASSERT_OK(base_file->Close());
+
+  std::unique_ptr<TableReader> base_reader;
+  NewTableReader(&base_reader);
+
+  std::string first_base_name = base_name_;
+  base_name_ = base_name_ + "second";
+  NewBaseFileWriter(&base_file);
```
yiwu-arbug: use another variable to avoid confusing the second base file with the first one?

src/table_builder_test.cc
wujy-cs: > I look at the code and seems the test table builder don't write to file with name as `blob_name_`. Am I missing something?

Since total size to write is larger than blob_file_target_size, it will write to multiple blob file with size about blob_file_target_size, blob_name_ is the first one.

src/table_builder_test.cc
wujy-cs: Ok. It's easier to reuse code using base_name_.

src/table_builder_test.cc
wujy-cs: @yiwu-arbug I modified the logic

PR #67, opened 2019-09-05
Fix metrics and gc mark

Connor1996:
Fix https://github.com/pingcap/titan/issues/65


src/db_impl.cc
```diff
@@ -223,6 +223,9 @@ Status TitanDBImpl::Open(const std::vector<TitanCFDescriptor>& descs,
   }
   if (!s.ok()) return s;
 
+  if (stats_.get()) {
```
yiwu-arbug: just curious, why's the change?

Connor1996: metrics issue

PR #68, opened 2019-09-06
Update BlobFileMeta format

Connor1996:
This is an extract of https://github.com/pingcap/titan/pull/51

Adding `smallest_key`, `largest_key`, `file_entries` and `file_level` in BlobFileMeta to support DeleteBlobFilesInRange and vtable design

### DIFF
--- a/src/blob_format.cc
+++ b/src/blob_format.cc
@@ -149,7 +149,7 @@ Status BlobFileMeta::DecodeFromLegacy(Slice* src) {
 }
 
 Status BlobFileMeta::DecodeFrom(Slice* src) {
-  if (!GetVarint64(src, &file_number_) || !GetVarint64(src, &file_size_) || 
+  if (!GetVarint64(src, &file_number_) || !GetVarint64(src, &file_size_) ||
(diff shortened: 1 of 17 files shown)


Connor1996:
PTAL @yiwu-arbug @wujy-cs

wujy-cs:
PTAL @yiwu-arbug

src/blob_format.cc
```diff
@@ -133,11 +133,36 @@ bool operator==(const BlobIndex& lhs, const BlobIndex& rhs) {
...
+  return Status::OK();
+}
+
+Status BlobFileMeta::DecodeFrom(Slice* src) {
+  if (!GetVarint64(src, &file_number_) || !GetVarint64(src, &file_size_) ||
+      !GetVarint64(src, &file_entries_) || !GetVarint32(src, &file_level_)) {
+    return Status::Corruption("BlobFileMeta decode failed");
+  }
+  Slice str;
+  if (GetLengthPrefixedSlice(src, &str)) {
```
yiwu-arbug: check key is non-zero size, or return corruption status. same below.

Connor1996:
@wujy-cs the check of the decoded key is not empty is not right, the decoded key may be empty.

PR #69, opened 2019-09-06
Rename version_set to blob_file_set

Connor1996:
Based on the current implementation, the `version_set` is not an appropriate name.


yiwu-arbug:
How about BlobFileSet? Since "blob" refer to each value.

PR #70, opened 2019-09-10
Persist gc stats via stats_cf of rocksdb

GlitterIsMe:
Persist stats of each blob_file by stats_cf of RocksDB, when OnFlushCompleted/OnCompactionCompleted/end of the GC, blob_file_num and discardable_size will be formatted and Write to stats_cf of RocksDB by Write, each modification in one flush/compaction/GC is persist atomically. Discardable size will be retrieved at DB Open.

This code is under debug now


Connor1996:
Implement it in a another way #130

PR #72, opened 2019-09-10
Prevent CF being dropped while GC is running

### LINKED ISSUE #71: Drop CF while GC is running causing background error
Found by db_stress #64. On `DropColumnFamilies` we should wait for any running GC job before proceed.

yiwu-arbug:
Summary:
If `DropColumnFamilies` is called while GC is running, GC can fail because the CF is gone. The GC job will then set background error which halts Titan as a whole. To prevent it, we let `DropColumnFamilies` wait till there no running GC before proceed, and let GC job wait for pending drop CF requests before start running. Fixes #71.

Also fix `DropColumnFamilies` mark obsolete file as obsolete again, which will cause assert failure.

Test Plan:
Run titandb_stress and the above issues don't reproduce.


src/db_impl.cc
```diff
@@ -397,11 +397,26 @@ Status TitanDBImpl::DropColumnFamilies(
...
   }
+  {
+    MutexLock l(&mutex_);
+    drop_cf_requests_++;
+
+    // Has to wait till no GC job is running before proceed, otherwise GC jobs
+    // can fail and set background error.
+    // TODO(yiwu): only wait for GC jobs of CFs being dropped.
+    while (bg_gc_running_ > 0) {
+      bg_cv_.Wait();
```
Connor1996: a column family is actually removed in `DestroyColumnFamilyHandle`, we should wait there instead

yiwu-arbug: We call `rocksdb::DropColumnFamilies()` of base db in `tiandb::DropColumnFamilies()`, after which GC will fail to write back to LSM, causing background error. We need to wait here.

yiwu-arbug:
> Can you add a unit test to reproduce?

yes, will do.

src/titan_db_test.cc
```diff
@@ -252,10 +254,17 @@ class TitanDBTest : public testing::Test {
...
+    {
+      MutexLock l(&db_impl_->mutex_);
+      db_impl_->bg_gc_scheduled_++;
+    }
     db_impl_->BackgroundCallGC();
-    while (db_impl_->bg_gc_scheduled_)
-      ;
+    {
+      MutexLock l(&db_impl_->mutex_);
+      while (db_impl_->bg_gc_scheduled_) {
```
Connor1996: mutex is held, `bg_gc_scheduled_` would not change

src/db_impl.cc
```diff
@@ -259,19 +259,15 @@ Status TitanDBImpl::Open(const std::vector<TitanCFDescriptor>& descs,
   s = blob_file_set_->Open(column_families);
   if (!s.ok()) return s;
 
-  static bool has_init_background_threads = false;
-  if (!has_init_background_threads) {
```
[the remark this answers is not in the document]
yiwu-arbug: totally no idea. I was from @DorianZheng 's initial version.

src/titan_db_test.cc
yiwu-arbug: cv_.Wait() implicitly unlock mutex. This is common pattern to use condition variables.

Connor1996: got it

src/db_impl.cc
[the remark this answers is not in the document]
yiwu-arbug: I think the intent was to increase the thread pool size by `max_background_gc`, but don't want to do it multiple times if `env` is shared by multiple db instance. I'm changing the logic to set thread pool size to `max_background_gc` instead.

src/db_impl.cc
```diff
@@ -261,13 +261,8 @@ Status TitanDBImpl::Open(const std::vector<TitanCFDescriptor>& descs,
...
   // Initialize GC thread pool.
   if (!db_options_.disable_background_gc && db_options_.max_background_gc > 0) {
-    auto bottom_pri_threads_num =
-        env_->GetBackgroundThreads(Env::Priority::BOTTOM);
-    env_->IncBackgroundThreadsIfNeeded(
-        db_options_.max_background_gc + bottom_pri_threads_num,
-        Env::Priority::BOTTOM);
-    assert(env_->GetBackgroundThreads(Env::Priority::BOTTOM) ==
-           bottom_pri_threads_num + db_options_.max_background_gc);
+    env_->IncBackgroundThreadsIfNeeded(db_options_.max_background_gc,
```
Connor1996: what if other works need more bottom threads

yiwu-arbug: then the other work should decide how many threads it needs (I agree this env API is awkward..)

PR #77, opened 2019-09-15
Live data based GC for level merge

wujy-cs:
This patch implemented live data based GC for level merge.

- Delete blob files which have (discardable size == file_size - file header - file footer)  in OnCompactionCompleted()
- Mark last two level blob files to merge in next compaction if discardable size reached GC threshold 

Using live data based GC instead of regular GC as long as level merge is enabled.


wujy-cs:
@yiwu-arbug PTAL

PS: Maybe I should use "records" instead of "entries" to represent KV number of blob file?

yiwu-arbug:
sorry for delay, but I'll review tomorrow.

src/table_builder.cc
```diff
@@ -220,7 +220,9 @@ TableProperties TitanTableBuilder::GetTableProperties() const {
 
 bool TitanTableBuilder::ShouldMerge(
     const std::shared_ptr<rocksdb::titandb::BlobFileMeta>& file) {
-  return file != nullptr && (int)file->file_level() < target_level_;
+  return file != nullptr &&
```
yiwu-arbug: `assert(cf_options_.level_merge)`

src/blob_format.h
```diff
@@ -217,6 +219,10 @@ class BlobFileMeta {
 
   void AddDiscardableSize(uint64_t _discardable_size);
   double GetDiscardableRatio() const;
+  void AddDiscardableEntries(uint64_t _discardable_entries);
+  bool Expired() {
```
yiwu-arbug: s/Expired/HasLiveData/ 

"expire" is usually about time.

src/blob_file_size_collector.cc
```diff
@@ -14,30 +14,36 @@ const std::string BlobFileSizeCollector::kPropertiesName =
...
 bool BlobFileSizeCollector::Encode(
-    const std::map<uint64_t, uint64_t>& blob_files_size, std::string* result) {
+    const std::map<uint64_t, BlobFileSize>& blob_files_size,
+    std::string* result) {
   PutVarint32(result, static_cast<uint32_t>(blob_files_size.size()));
   for (const auto& bfs : blob_files_size) {
     PutVarint64(result, bfs.first);
-    PutVarint64(result, bfs.second);
+    PutVarint64(result, bfs.second.size);
+    PutVarint64(result, bfs.second.entries);
```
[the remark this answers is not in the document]
wujy-cs: @yiwu-arbug Okay. Please take another look.

src/blob_gc_job_test.cc
```diff
@@ -464,6 +464,58 @@ TEST_F(BlobGCJobTest, DeleteFilesInRange) {
   DestroyDB();
 }
 
+TEST_F(BlobGCJobTest, LevelMergeGC) {
```
yiwu-arbug: I think you need to disable purge_obsolete_files thread for the test, otherwise it can be flaky. How about change `StartBackgroundTasks` to not start purge_obsolete_files thread if `purge_obsolete_files_period_sec=0`, and set the option to 0 for this test?

PR #78, opened 2019-09-16
Fix new CF missing BlobFileSizeCollector

yiwu-arbug:
Summary:
On `CreateColumnFamilies` we should set `BlobFileSizeCollectorFactory`. Fixing it.

Test Plan:
added test


yiwu-arbug:
seems CI finds a data race case. will fix.

PR #88, opened 2019-09-23
Reset gc mark after sampling

### LINKED ISSUE #87: Should update gc mark after sampling
if gc mark is not updated, the picker will always select the same blob files and do not make progress.

Connor1996:
Reset gc mark after sampling, Fix #87

### DIFF
--- a/src/blob_file_set.cc
+++ b/src/blob_file_set.cc
@@ -94,6 +94,12 @@ Status BlobFileSet::Recover() {
                    "Next blob file number is %" PRIu64 ".", next_file_number);
   }
 
+  // Make sure perform gc on all files at the beginning
+  MarkAllFilesForGC();
+  for (auto& cf : column_families_) {
+    cf.second->ComputeGCScore();
+  }
+
   auto new_manifest_file_number = NewFileNumber();
   s = OpenManifest(new_manifest_file_number);
   if (!s.ok()) return s;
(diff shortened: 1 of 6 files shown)


src/blob_file_set.cc
```diff
@@ -94,6 +94,12 @@ Status BlobFileSet::Recover() {
                    "Next blob file number is %" PRIu64 ".", next_file_number);
   }
 
+  // Make sure perform gc on all files at the beginning
+  MarkAllFilesForGC();
```
yiwu-arbug: Just curious. Why the call needs to move up here?

src/blob_gc_picker.cc
```diff
@@ -71,8 +67,17 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
                   "got batch size %" PRIu64 ", estimate output %" PRIu64
                   " bytes",
                   batch_size, estimate_output_size);
-  if (blob_files.empty() || batch_size < cf_options_.min_gc_batch_size)
+  if (blob_files.empty() || batch_size < cf_options_.min_gc_batch_size) {
+    return nullptr;
+  }
+  // if there is only one small file to merge, no need to perform
+  if (blob_files.size() == 1 &&
```
yiwu-arbug: I think we shouldn't merge small files if the end result is still less than `merge_small_file_threshold`, but let's do it next time.

src/blob_file_set.cc
Connor1996: oh, no need actually. I forget to revert it.

PR #92, opened 2019-09-29
Implement basic range merge to maintain last level sorted run number

wujy-cs:
With level merge enabled, we expect there are no more than 10 sorted runs of blob files in both of last two levels. But 

- Since we force using level_compaction_dynamic_level_bytes, last level blob files won't be merged again, sorted run number in last level will increase infinitely.
- Some hot range may contain more sorted runs.

This patch implement range merge:

- Check sorted run number of compaction range after last level compaction
- Mark blob files to merge in next compaction if sorted run number > max_sorted_runs


include/titan/options.h
```diff
@@ -137,6 +137,24 @@ struct TitanCFOptions : public ColumnFamilyOptions {
...
+  // of blob files in both of last two levels. But since last level blob files
+  // won't be merged again, sorted runs in last level will increase infinitely.
+  //
+  // With this feature enabled, Titan will check sorted runs of compaction range
+  // after each last level compaction and mark related blob files if there are
+  // too many. These marked blob files will be merged to a new sorted run in
+  // next compaction.
+  //
+  // Default: false
+  bool range_merge{false};
```
Connor1996: why this is optional? Is there any circumstances where no need to do range merge?

src/db_impl.cc
```diff
@@ -837,6 +837,46 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
...
+    }
+    sorted_run_blobs.push_back(cnt);
+  }
+
+  // There might be several sorted runs consist of few wide range blob files.
+  // Ignore these sorted runs since it has almost no impact to scan performance.
+  std::sort(sorted_run_blobs.begin(), sorted_run_blobs.end());
+  int i = 0, ignore = num_files * 0.1;
+  while (sorted_run_blobs[i] < ignore) {
+    ignore -= sorted_run_blobs[i++];
```
Connor1996: why `ignore` should minus the ignored runs?

include/titan/options.h
wujy-cs: Range merge is to improve short range scan performance, at circumstances that most query is long range scan, range merge is unnecessary.

src/db_impl.cc
wujy-cs: sorted_run_blobs is number of blob files in each sorted run, we ignore 10% files since some sorted runs have only very few files which has no much impacts to scan performance.

src/db_impl.cc
```diff
@@ -837,6 +837,46 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
...
+        remained_files.erase(iter++);
+      } else {
+        iter++;
+      }
+    }
+    sorted_run_blobs.push_back(cnt);
+  }
+
+  // There might be several sorted runs consist of few wide range blob files.
+  // Ignore these sorted runs since it has almost no impact to scan performance.
```
yiwu-arbug: I don't quite get this optimization, and don't  know to reason about the 10% number. How about make the implementation simpler and skip this optimization?

src/db_impl.cc
```diff
@@ -1146,6 +1200,11 @@ void TitanDBImpl::OnCompactionCompleted(
     // data based GC, so we don't need to trigger regular GC anymore
     if (cf_options.level_merge) {
       blob_file_set_->LogAndApply(edit);
+      if (count_sr && CountSortedRuns(files) > cf_options.max_sorted_runs) {
```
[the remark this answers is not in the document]
wujy-cs: CountSortedRuns() counts all blob files of this compaction so we need mark all blob files. ( eg. the granularity is entire compaction range)

src/db_impl.cc
wujy-cs: Okay, we can skip this. This optimization is from my previous experiment, that some sorted runs consist of only 10% blob files with this estimate method.

src/db_impl.cc
```diff
@@ -1132,12 +1182,16 @@ void TitanDBImpl::OnCompactionCompleted(
         if (file->NoLiveData()) {
           edit.DeleteBlobFile(file->file_number(),
                               db_impl_->GetLatestSequenceNumber());
+          continue;
         } else if (static_cast<int>(file->file_level()) >=
                        cf_options.num_levels - 2 &&
                    file->GetDiscardableRatio() >
                        cf_options.blob_file_discardable_ratio) {
           file->FileStateTransit(BlobFileMeta::FileEvent::kNeedMerge);
```
Connor1996: It's not related to this PR, but I have a question on how to make sure the file can be deleted by level range.
Cause `TableBuilder::Add` gives no chance for deleted entries, a blob file with a lot of deleted entries may never be checked.

src/db_impl.cc
Connor1996: oh, I misunderstand the code.

src/db_impl.cc
wujy-cs: @Connor1996 
Sorry for delay.
Live data is not checked by TableBuilder::Add but BlobFileSizeCollector in OnCompactionCompleted(), it compares SSTable referenced value size before/after compaction, so deleted entries will be checked while compacting the deletion record with previous record.

src/db_impl.cc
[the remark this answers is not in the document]
yiwu-arbug: I mean, I think we can change `CountSortedRuns()` algorithm to have finer granularity.

src/db_impl.cc
```diff
@@ -837,6 +837,46 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
   return s;
 }
 
+int TitanDBImpl::CountSortedRuns(
```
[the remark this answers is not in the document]
wujy-cs: @yiwu-arbug 
Sorry for delay!
You mean maintain ends of each sorted run and compare every file with them? This has O(n*m) complexity, maybe it's not faster than current method.

src/db_impl.cc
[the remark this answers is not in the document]
wujy-cs: @yiwu-arbug I think this granularity is fine enough for most cases. A compaction includes 11 sstables at most, say we use 8MB sstable and 32MB blob file, value is larger than index about 32 times. So this compaciton includes about 80 blob files, and 8 blob files in a sorted run in average if there is 10 sorted runs.

But sure we can partition blob files to multiple subset and count sorted runs for each of them.

src/db_impl.cc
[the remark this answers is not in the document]
yiwu-arbug: Does something like this work? https://stackoverflow.com/a/2245026/11014942

wujy-cs:
@yiwu-arbug 
I optimized the algorithm, please take another look.

src/db_impl.cc
```diff
@@ -837,6 +837,50 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
...
+
+  // store and sort both ends of blob files to count sorted runs
+  std::vector<std::pair<BlobFileMeta*, bool /* is smallest end? */>> blob_ends;
+  for (const auto& file : files) {
+    blob_ends.emplace_back(std::make_pair(file.get(), true));
+    blob_ends.emplace_back(std::make_pair(file.get(), false));
+  }
+  auto blob_ends_cmp = [](const std::pair<BlobFileMeta*, bool>& end1,
+                          const std::pair<BlobFileMeta*, bool>& end2) {
+    std::string key1 = std::move(end1.second ? end1.first->smallest_key()
```
yiwu-arbug: you cannot move the content of smallest_key() or largest_key(), or the next call of comparison on the file will find the key to be empty. Just use reference
```
std::string& key1 = end1.second ? end.first->smallest_key() : end->first->largest_key();
```

src/db_impl.cc
```diff
@@ -837,6 +837,50 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
...
+                                             : end1.first->largest_key());
+    std::string key2 = std::move(end2.second ? end2.first->smallest_key()
+                                             : end2.first->largest_key());
+    return key1.compare(key2) < 0;
+  };
+  std::sort(blob_ends.begin(), blob_ends.end(), blob_ends_cmp);
+
+  int cur = 0, start = 0, size = blob_ends.size();
```
yiwu-arbug: style: one variable per line.

src/db_impl.cc
```diff
@@ -837,6 +837,50 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
...
+    blob_ends.emplace_back(std::make_pair(file.get(), true));
+    blob_ends.emplace_back(std::make_pair(file.get(), false));
+  }
+  auto blob_ends_cmp = [](const std::pair<BlobFileMeta*, bool>& end1,
+                          const std::pair<BlobFileMeta*, bool>& end2) {
+    std::string key1 = std::move(end1.second ? end1.first->smallest_key()
+                                             : end1.first->largest_key());
+    std::string key2 = std::move(end2.second ? end2.first->smallest_key()
+                                             : end2.first->largest_key());
+    return key1.compare(key2) < 0;
```
yiwu-arbug: when the key being the same, order largest_keys before smallest_keys.

src/db_impl.cc
```diff
@@ -837,6 +837,50 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
...
+  std::sort(blob_ends.begin(), blob_ends.end(), blob_ends_cmp);
+
+  int cur = 0, start = 0, size = blob_ends.size();
+  for (int i = 0; i < size; i++) {
+    if (blob_ends[i].second) {
+      ++cur;
+    } else {
+      --cur;
+      // reach max sorted run in [start, i], mark these blob files to merge
+      if (cur == max_sorted_runs) {
```
yiwu-arbug: `cur >= max_sorted_runs` to be safer.

src/db_impl.cc
wujy-cs: > you cannot move the content of smallest_key() or largest_key(), or the next call of comparison on the file will find the key to be empty. Just use reference
> 
> ```
> std::string& key1 = end1.second ? end.first->smallest_key() : end->first->largest_key();
> ```

ok. But smallest_key() returned a new constructed string, I thought it won't affect filemeta?

src/db_impl.cc
```diff
@@ -837,6 +837,50 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
   return s;
 }
 
+void TitanDBImpl::MaybeScheduleRangeMerge(
```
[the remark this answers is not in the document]
wujy-cs: we already have one, see BlobGCJobTest->RangeMerge

src/db_impl.cc
```diff
@@ -837,6 +837,50 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
...
+    return key1.compare(key2) < 0;
+  };
+  std::sort(blob_ends.begin(), blob_ends.end(), blob_ends_cmp);
+
+  int cur = 0, start = 0, size = blob_ends.size();
+  for (int i = 0; i < size; i++) {
+    if (blob_ends[i].second) {
+      ++cur;
+    } else {
+      --cur;
```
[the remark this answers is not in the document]
wujy-cs: right

src/blob_gc_job_test.cc
```diff
@@ -580,7 +588,97 @@ TEST_F(BlobGCJobTest, LevelMergeGC) {
   DestroyDB();
 }
 
+TEST_F(BlobGCJobTest, RangeMergeScheduler) {
+  NewDB();
+  int max_sorted_run = 2;
+  std::vector<std::shared_ptr<BlobFileMeta>> files;
+
+  // one sorted run, no file will be mark
```
Connor1996:
```suggestion
  // one sorted run, no file will be marked
```

src/blob_gc_job_test.cc
```diff
@@ -580,7 +588,97 @@ TEST_F(BlobGCJobTest, LevelMergeGC) {
...
+                                       std::string('a' + i * 2 + 1, 1));
+    file->FileStateTransit(BlobFileMeta::FileEvent::kDbRestart);
+    files.emplace_back(file);
+  }
+  ScheduleRangeMerge(files, max_sorted_run);
+  for (const auto& file : files) {
+    ASSERT_EQ(file->file_state(), BlobFileMeta::FileState::kNormal);
+  }
+
+  // one more sorted run for 5 files, only those files will be marked
```
Connor1996: please add illustration like this
```
runs1:[a_b) [d_____________n) [o___q)
runs2:          [e______g)
```
so we can easily get the layout

src/blob_gc_job_test.cc
```diff
@@ -580,7 +588,101 @@ TEST_F(BlobGCJobTest, LevelMergeGC) {
...
+                                       std::string('a' + i * 2 + 1, 1));
+    file->FileStateTransit(BlobFileMeta::FileEvent::kDbRestart);
+    files.emplace_back(file);
+  }
+  ScheduleRangeMerge(files, max_sorted_run);
+  for (const auto& file : files) {
+    ASSERT_EQ(file->file_state(), BlobFileMeta::FileState::kNormal);
+  }
+
+  // one more sorted run for 6 files, only those files will be marked
```
Connor1996: please add a case that have one's end == other's start;
please add a case like this
```
run 1: [a             f]
run 2:    [b,      e]
run 3:      [c,d]
```
please add a case like this
```
run 1: [a, b] [c, d] [e, f] ... [n, o] [p, q] [r, s]
run 2:        [c, d] [e, f]            [p, q] [r, s]
```

src/db_impl.cc
```diff
@@ -847,6 +847,54 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
...
+  }
+  auto blob_ends_cmp = [](const std::pair<BlobFileMeta*, bool>& end1,
+                          const std::pair<BlobFileMeta*, bool>& end2) {
+    const std::string& key1 =
+        end1.second ? end1.first->smallest_key() : end1.first->largest_key();
+    const std::string& key2 =
+        end2.second ? end2.first->smallest_key() : end2.first->largest_key();
+    int cmp = key1.compare(key2);
+    // when the key being the same, order largest_key before smallest_key
+    return cmp == 0 ? end2.second : cmp < 0;
```
yiwu-arbug: `return cmp == 0 ? (!end1.second && end2.second) : cmp < 0;`

src/db_impl.cc
```diff
@@ -847,6 +847,54 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
...
+                BlobFileMeta::FileEvent::kNeedMerge);
+          }
+          ++start;
+        }
+      } else if (cur == 1) {
+        start = i + 1;
+      }
+      --cur;
+    }
+  }
```
yiwu-arbug: let say there are files with range [a,h], [b,c], [d,g], [e,f] and max_sorted_runs = 3, The current algorithm will mark [b,c] to merge, while it doesn't belong to a 3-sorted-run range.

src/db_impl.cc
wujy-cs: @yiwu-arbug 
I think it's ok. If end2 is largest (!end2.second) then we can order it before end1 no matter what end1 is.

src/db_impl.cc
wujy-cs: @yiwu-arbug 
Yes. I updated the algorithm to resolve this.

src/db_impl.cc
```diff
@@ -847,6 +847,48 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
...
+  auto blob_ends_cmp = [](const std::pair<BlobFileMeta*, bool>& end1,
+                          const std::pair<BlobFileMeta*, bool>& end2) {
+    const std::string& key1 =
+        end1.second ? end1.first->smallest_key() : end1.first->largest_key();
+    const std::string& key2 =
+        end2.second ? end2.first->smallest_key() : end2.first->largest_key();
+    int cmp = key1.compare(key2);
+    // when the key being the same, order largest_key before smallest_key
+    return cmp == 0 ? !end1.second && end2.second : cmp < 0;
```
yiwu-arbug: nit: add bracket for readability? i.e. `(cmp == 0) ? (!end1.second && end2.second) : (cmp < 0);`

src/db_impl.cc
```diff
@@ -847,6 +847,48 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
...
+  int size = blob_ends.size();
+  std::unordered_map<BlobFileMeta*, int> tmp;
+  for (int i = 0; i < size; i++) {
+    if (blob_ends[i].second) {
+      ++cur_add;
+      tmp[blob_ends[i].first] = cur_remove;
+    } else {
+      ++cur_remove;
+      auto record = tmp.find(blob_ends[i].first);
+      if (cur_add - record->second >= max_sorted_runs) {
```
Connor1996:
```suggestion
      if (cur_add - record->second > max_sorted_runs) {
```
should mark only when exceed the threshold, which makes the behavior more consistent with the literal meaning of `max_sorted_runs`

Issue #93, opened 2019-10-08
concurrent flush cause data loss after GC


yiwu-arbug:
Titan use event listener to hook `OnFlushCompleted`. When the event triggered, Titan assume corresponding memtable has been converted to SST, and mark blob file generated from the flush as normal file, so that GC can pick the file up. However, the assumption is not correct. If there are concurrent flush happened, `OnFlushCompleted` can fire before memtable being converted to SST (https://github.com/facebook/rocksdb/issues/5892). If GC pick up the blob file before memtable flush finished, it can receive is_blob_index=false for all keys in the blob file (since they are still in memtable), and drop all data in the file by mistake, causing data loss.

Reproduced with unit test: https://github.com/yiwu-arbug/titan/commit/5172201d64b84b8b9bf9f32d50d54dfaf8433899

PR #96, opened 2019-10-15
Temp fix for data loss caused by concurrent flush

yiwu-arbug:
Summary:
Temporary fix for #93. Fixing it by waiting until no flush is ongoing before starting a GC job. The drawback is GC can starve if flush runs so frequent that there's no gap between flush jobs.

Test Plan:
See the new test. The test will fail with "Corruption: Missing blob file" error before the fix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,11 +5,11 @@ enable_language(C)
 find_package(Git)
 
 if (NOT ROCKSDB_GIT_REPO)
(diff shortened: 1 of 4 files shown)


yiwu-arbug:
Need to merge https://github.com/pingcap/rocksdb/pull/126 before merging this one.

Connor1996:
Please merge latest master

PR #101, opened 2019-10-31
Add log for blob file range

Connor1996:


src/blob_gc_job.cc
```diff
@@ -585,8 +585,10 @@ Status BlobGCJob::DeleteInputBlobFiles() {
   edit.SetColumnFamilyID(blob_gc_->column_family_handle()->GetID());
   for (const auto& file : blob_gc_->sampled_inputs()) {
     ROCKS_LOG_INFO(db_options_.info_log,
-                   "Titan add obsolete file [%" PRIu64 "]",
-                   file->file_number());
+                   "Titan add obsolete file [%" PRIu64 "] range [%s, %s]",
+                   file->file_number(),
+                   file->smallest_key().ToString(true).c_str(),
```
yiwu-arbug: Printing key to log may not be a good idea. Like, the key may contain sensitive data and user may not want it leak to logs.

Connor1996: should we remove it？Though TiKV does print range in logs🤣

PR #104, opened 2019-11-05
Fix algorithm for counting runs

Connor1996:
make the algorithm simpler and add some cases where origin algorithm fails

### DIFF
--- a/src/db_impl.cc
+++ b/src/db_impl.cc
@@ -878,17 +878,17 @@ void TitanDBImpl::MarkFileIfNeedMerge(
   };
   std::sort(blob_ends.begin(), blob_ends.end(), blob_ends_cmp);
 
-  std::unordered_set<BlobFileMeta*> tmp;
+  std::unordered_set<BlobFileMeta*> set;
   for (auto& end : blob_ends) {
     if (end.second) {
-      tmp.insert(end.first);
-      if (tmp.size() > (size_t)max_sorted_runs) {
-        for (auto file : tmp) {
+      set.insert(end.first);
+      if (set.size() > static_cast<size_t>(max_sorted_runs)) {
+        for (auto file : set) {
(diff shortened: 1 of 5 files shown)


tools/db_bench_tool.cc
```diff
@@ -760,6 +760,10 @@ DEFINE_uint64(blob_db_min_blob_size, 0,
 // Titan Options
 DEFINE_bool(use_titan, true, "Open a Titan instance.");
 
+DEFINE_bool(titan_level_merge, true, "Enable Titan level merge.");
```
yiwu-arbug: default to false?

src/db_impl.cc
```diff
@@ -878,20 +878,17 @@ void TitanDBImpl::MarkFileIfNeedMerge(
...
-  for (int i = 0; i < size; i++) {
-    if (blob_ends[i].second) {
-      ++cur_add;
-      tmp[blob_ends[i].first] = cur_remove;
-    } else {
-      ++cur_remove;
-      auto record = tmp.find(blob_ends[i].first);
-      if (cur_add - record->second > max_sorted_runs) {
-        record->first->FileStateTransit(BlobFileMeta::FileEvent::kNeedMerge);
+  std::unordered_set<BlobFileMeta*> tmp;
```
yiwu-arbug: nit: give it a better name than `tmp`?

src/db_impl.cc
```diff
@@ -878,20 +878,17 @@ void TitanDBImpl::MarkFileIfNeedMerge(
...
+  for (auto& end : blob_ends) {
+    if (end.second) {
+      tmp.insert(end.first);
+      if (tmp.size() > (size_t)max_sorted_runs) {
```
yiwu-arbug: nit: use c++ style cast, i.e. `static_cast<size_t>(...)`

Issue #108, opened 2019-11-15
PCP-23-1: Update blob file format to accommodate zstd dictionary


yiwu-arbug:
This is subtask of https://github.com/tikv/tikv/issues/5743.

We need to add a new version to blob file format. If version = 2, it means the blob file come with a zstd dictionary and the dictionary is stored in the meta section (pointed by meta index handle in the footer). Update `BlobFileHeader` and `BlobFileFooter` and their comments accordingly.

Also update `BlobEncoder` and `BlobDecoder` to allow passing in a dictionary and use it to  compress/decompress. Add unit test for it.

This should be a good warmup task for you to get familiar with Titan development.

yiwu-arbug:
@foeb

PR #109, opened 2019-11-17
Update blob file format for ZSTD dictionary compression

foeb:
#108

### DIFF
--- a/src/blob_file_iterator.cc
+++ b/src/blob_file_iterator.cc
@@ -18,11 +18,11 @@ BlobFileIterator::~BlobFileIterator() {}
 
 bool BlobFileIterator::Init() {
   Slice slice;
-  char header_buf[BlobFileHeader::kEncodedLength];
+  char header_buf[BlobFileHeader::kMaxEncodedLength];
   // With for_compaction=true, rate_limiter is enabled. Since BlobFileIterator
   // is only used for GC, we always set for_compaction to true.
-  status_ = file_->Read(0, BlobFileHeader::kEncodedLength, &slice, header_buf,
-                        true /*for_compaction*/);
+  status_ = file_->Read(0, BlobFileHeader::kMaxEncodedLength, &slice,
+                        header_buf, true /*for_compaction*/);
   if (!status_.ok()) {
     return false;
   }
@@ -31,6 +31,11 @@ bool BlobFileIterator::Init() {
   if (!status_.ok()) {
     return false;
   }
+
+  header_size_ = blob_file_header.version == BlobFileHeader::kVersion1
+                     ? BlobFileHeader::kMinEncodedLength
+                     : BlobFileHeader::kMaxEncodedLength;
+
   char footer_buf[BlobFileFooter::kEncodedLength];
   // With for_compaction=true, rate_limiter is enabled. Since BlobFileIterator
   // is only used for GC, we always set for_compaction to true.
@@ -42,7 +47,7 @@ bool BlobFileIterator::Init() {
   status_ = blob_file_footer.DecodeFrom(&slice);
   end_of_blob_record_ = file_size_ - BlobFileFooter::kEncodedLength -
                         blob_file_footer.meta_index_handle.size();
-  assert(end_of_blob_record_ > BlobFileHeader::kEncodedLength);
+  assert(end_of_blob_record_ > BlobFileHeader::kMinEncodedLength);
   init_ = true;
   return true;
 }
@@ -50,7 +55,7 @@ bool BlobFileIterator::Init() {
 void BlobFileIterator::SeekToFirst() {
   if (!init_ && !Init()) return;
   status_ = Status::OK();
-  iterate_offset_ = BlobFileHeader::kEncodedLength;
+  iterate_offset_ = header_size_;
   PrefetchAndGet();
 }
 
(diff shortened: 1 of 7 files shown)


src/blob_format.h
```diff
@@ -253,10 +260,14 @@ class BlobFileMeta {
...
 // The header is mean to be compatible with header of BlobDB blob files, except
 // we use a different magic number.
+//
+// If the version is 2, then the meta section includes a (possibly empty)
+// uncompression dictionary.
 struct BlobFileHeader {
   // The first 32bits from $(echo titandb/blob | sha1sum).
   static const uint32_t kHeaderMagicNumber = 0x2be0a614ul;
   static const uint32_t kVersion1 = 1;
+  static const uint32_t kVersion2 = 2;
```
yiwu-arbug: After a second thought, I think we should leave a 4 bytes padding in the file header, and use 1 bit as a flag to mean whether the file come with a dictionary. The rest of bits can leave room for future extension. I would like the have v2 mean "have extra flags in the header", not "have dictionary".

foeb: Ok, I added a flags field. Right now it does nothing, but is serialized/deserialized when the version is 2 and complains if any unusual bits are set. I also didn't create an enum for the flags (since there's only one), but that would be an easy change if you'd prefer that. Let me know if I missed anything.

foeb:
@Connor1996 
Ok, I updated the description. Does that look right?

yiwu-arbug:
@foeb Sorry, I have some comments that I failed to send out earlier. Mind address them as well? Thanks.

src/blob_format.cc
```diff
@@ -280,6 +284,11 @@ Status BlobFileHeader::DecodeFrom(Slice* src) {
       (version != kVersion1 && version != kVersion2)) {
     return Status::Corruption("Blob file header version missing or invalid.");
   }
+  if (version == 2) {
+    if (!GetFixed32(src, &flags) || flags & ~kHasUncompressionDictionary) {
```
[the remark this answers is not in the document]
foeb: I might've done the logic wrong, but it was supposed to check that no other flags were set. I'll remove it anyway because it doesn't look right.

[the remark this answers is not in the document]
yiwu-arbug: I misread the code. Looking good then. But let's add a inline comment for it?

src/blob_format.h
```diff
@@ -250,27 +250,36 @@ class BlobFileMeta {
...
   static const uint32_t kHeaderMagicNumber = 0x2be0a614ul;
   static const uint32_t kVersion1 = 1;
   static const uint32_t kVersion2 = 2;
   static const uint64_t kEncodedLength = 4 + 4;
 
+  // Flags:
+  static const uint32_t kHasUncompressionDictionary = 1 << 0;
+
   uint32_t version = kVersion1;
+  uint32_t flags = 0;
```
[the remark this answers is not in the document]
foeb: I was able to do this, but it ended up breaking a couple `BlobGCJobTest`s by including the extra 4 bytes in the header. I haven't found a reason why it break in titan's code, yet. Is it possible rocksdb assumes the header is 8 bytes?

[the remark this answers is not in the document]
yiwu-arbug: Not rocksdb, but somewhere else in titan (at least in `BlobFileIterator`) it makes the assumption. We should update those call-sites accordingly.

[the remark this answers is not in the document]
foeb: I looked again, and all I needed to do was update `kBlobHeaderSize`. :woman_facepalming: 
I think everything is taken care of now

src/blob_format.h
```diff
@@ -34,7 +34,7 @@ namespace titandb {
 //    | Fixed32 | Fixed32 |    char     |
 //    +---------+---------+-------------+
 //
-const uint64_t kBlobHeaderSize = 8;
+const uint64_t kBlobHeaderSize = 12;
```
yiwu-arbug: s/kBlobHeaderSize/kMaxBlobHeaderSize/

src/blob_gc_job.cc
```diff
@@ -184,12 +184,12 @@ Status BlobGCJob::DoSample(const BlobFileMeta* file, bool* selected) {
   if (*selected) return Status::OK();
 
   // TODO: add do sample count metrics
-  auto records_size = file->file_size() - BlobFileHeader::kEncodedLength -
+  auto records_size = file->file_size() - BlobFileHeader::kMaxEncodedLength -
```
yiwu-arbug: You need to add a `header_size()` method to `FileMetaData` and return result depend on header version. 

Actually leave an inline comment here to note `records_size` is not accurate is also fine. The `DoSample` method is being removed soon.

src/blob_format.cc
```diff
@@ -285,6 +285,7 @@ Status BlobFileHeader::DecodeFrom(Slice* src) {
     return Status::Corruption("Blob file header version missing or invalid.");
   }
   if (version == 2) {
```
yiwu-arbug: nit: version == kVersion2

src/blob_file_iterator.cc
```diff
@@ -31,6 +31,11 @@ bool BlobFileIterator::Init() {
   if (!status_.ok()) {
     return false;
   }
+
+  header_size_ = blob_file_header.version == BlobFileHeader::kVersion1
```
yiwu-arbug: let's add a `size()` method to `BlobFileHeader` and move the logic there.

yiwu-arbug:
Almost there!

yiwu-arbug:
@foeb Merged. Awesome work!

PR #111, opened 2019-11-18
Fix status overrided by mistake

Connor1996:
When `InstallOutputBlobFiles`, if there is an error of finishing a blob file, Titan would delete the blob file physically. But now, the non-okay-status is overridden by the okay-result of deleting files. So in the outer place, it would rewrite blob index to LSM as usual while the blob file is actually 
deleted.

### DIFF
--- a/src/blob_gc_job.cc
+++ b/src/blob_gc_job.cc
@@ -520,12 +520,12 @@ Status BlobGCJob::InstallOutputBlobFiles() {
       to_delete_files.append(std::to_string(builder.first->GetNumber()));
       handles.emplace_back(std::move(builder.first));
     }
-    ROCKS_LOG_BUFFER(
-        log_buffer_,
-        "[%s] InstallOutputBlobFiles failed. Delete GC output files: %s",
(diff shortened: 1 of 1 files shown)


src/blob_gc_job.cc
```diff
@@ -520,12 +520,12 @@ Status BlobGCJob::InstallOutputBlobFiles() {
       to_delete_files.append(std::to_string(builder.first->GetNumber()));
       handles.emplace_back(std::move(builder.first));
     }
-    ROCKS_LOG_BUFFER(
-        log_buffer_,
-        "[%s] InstallOutputBlobFiles failed. Delete GC output files: %s",
-        blob_gc_->column_family_handle()->GetName().c_str(),
-        to_delete_files.c_str());
-    s = blob_file_manager_->BatchDeleteFiles(handles);
+    Status status = blob_file_manager_->BatchDeleteFiles(handles);
```
yiwu-arbug: Add inline comment to note we should not set `s` here? Also print warning log if delete file failed?

PR #114, opened 2019-11-20
TitanDBImpl::Open only open base DB once

yiwu-arbug:
Summary:
Refactor `TitanDBImpl::Open` to open base DB only once. Previously we open the DB twice - the first time only to obtain `cf_id` for each CF, and then initialize internals before open base DB for real. The patch change to open base DB only once, initialize internals after that. Before `TitanDBImpl` is initialized, `BaseDbListener` will be no-op and `TableBuilder` will not output blob files.

Test Plan:
Added new tests


src/db_impl.cc
```diff
@@ -139,6 +139,7 @@ TitanDBImpl::TitanDBImpl(const TitanDBOptions& options,
   if (db_options_.statistics != nullptr) {
     stats_.reset(new TitanStats(db_options_.statistics.get()));
   }
+  blob_file_set_.reset(new BlobFileSet(db_options_, stats_.get()));
```
Connor1996: info log is initialized after `CreateLoggerFromOptions`, please make sure BlobFileSet is created after log initialization.

src/db_impl.cc
```diff
@@ -175,114 +176,140 @@ Status TitanDBImpl::ValidateOptions(
...
+  if (stats_ != nullptr) {
+    stats_->Initialize(column_families);
+  }
+  s = blob_file_set_->Open(column_families);
+  if (!s.ok()) {
+    return s;
+  }
+  TEST_SYNC_POINT_CALLBACK("TitanDBImpl::OpenImpl:BeforeInitialized", db_);
+  // Initialization done.
+  initialized_ = true;
```
Connor1996: seems `initialized_` should be set in background later?

src/db_impl.cc
yiwu-arbug: good catch

src/db_impl.cc
yiwu-arbug: No. After `initialized_=true` is set, compaction and flush can generate blob files. After the GC stats patch (which I'm working on), GC stats will be initialize after this point, in background, and after that GC job will start.

src/db_impl.cc
```diff
@@ -1015,6 +1045,11 @@ bool TitanDBImpl::GetIntProperty(ColumnFamilyHandle* column_family,
 }
 
 void TitanDBImpl::OnFlushCompleted(const FlushJobInfo& flush_job_info) {
+  TEST_SYNC_POINT("TitanDBImpl::OnFlushCompleted:Begin");
+  if (!initialized()) {
```
Connor1996: why check twice in basedb listener and here?

yiwu-arbug: just to be safe.

yiwu-arbug:
@Connor1996 any more comments?

src/db_impl.cc
```diff
@@ -175,114 +175,144 @@ Status TitanDBImpl::ValidateOptions(
...
-    s = db_->Close();
-    delete db_;
-    db_ = nullptr;
-  }
-  if (!s.ok()) return s;
-
-  if (stats_.get()) {
-    stats_->Initialize(column_families);
+  // Setup options.
+  db_options_.listeners.emplace_back(std::make_shared<BaseDbListener>(this));
```
Connor1996: seems this should set before `blob_file_set.reset` too.
To avoid other similar issues, maybe we should pass `db_option` when `BlobFileSet::Open`

yiwu-arbug: why? does blob_file_set use the listeners?

Connor1996: Oh, I misread the code.

PR #116, opened 2019-11-21
Use prefetcher for level merge

wujy-cs:
since level merge will read blob files sequentially, we should use prefetcher to accelerate merging.

experiment result:

Load 300GB (average value size 1KB) on Huawei NVME device, time to read blob file reduced about 50%.

### DIFF
--- a/src/table_builder.cc
+++ b/src/table_builder.cc
@@ -34,12 +34,7 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
     }
 
     BlobRecord record;
-    PinnableSlice buffer;
-
-    auto storage = blob_storage_.lock();
-    assert(storage != nullptr);
-    ReadOptions options;  // dummy option
-    Status get_status = storage->Get(options, index, &record, &buffer);
+    Status get_status = GetBlobRecord(index, &record);
     UpdateIOBytes(prev_bytes_read, prev_bytes_written, &io_bytes_read_,
                   &io_bytes_written_);
     if (get_status.ok()) {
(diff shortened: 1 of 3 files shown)


src/table_builder.h
```diff
@@ -69,6 +69,8 @@ class TitanTableBuilder : public TableBuilder {
   std::vector<
       std::pair<std::shared_ptr<BlobFileMeta>, std::unique_ptr<BlobFileHandle>>>
       finished_blobs_;
+  std::unordered_map<uint64_t, std::unique_ptr<BlobFilePrefetcher>>
+      merging_blobs_;
```
yiwu-arbug: nit: s/merging_blobs_/file_prefetchers_/

wujy-cs:
> LGTM.
> nit: Maybe we can leverage Prefetch for fallback too.

wujy-cs:
> LGTM.
> nit: Maybe we can leverage Prefetch for fallback too.

PTAL

src/table_builder.cc
```diff
@@ -266,5 +261,29 @@ void TitanTableBuilder::UpdateInternalOpStats() {
   }
 }
 
+Status TitanTableBuilder::GetBlobRecord(const BlobIndex& index,
+                                        BlobRecord* record) {
+  auto storage = blob_storage_.lock();
```
Connor1996: no need to call it everytime

src/table_builder.cc
```diff
@@ -84,9 +79,9 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
     auto blob_file = storage->FindFile(index.file_number).lock();
     if (ShouldMerge(blob_file)) {
       BlobRecord record;
-      PinnableSlice buffer;
-      Status s = storage->Get(ReadOptions(), index, &record, &buffer);
-      if (s.ok()) {
+      Status get_status = GetBlobRecord(index, &record);
+
+      if (get_status.ok()) {
```
Connor1996: if not okay, maybe we should set `status_`  and return, or it will neglect error silently.

src/table_builder.cc
wujy-cs: How. Is it safe to hold a shared_ptr in TitanTableBuilder?

src/table_builder.cc
wujy-cs: If not ok it will write original key-value_index to base_builder instead of doing level_merge

yiwu-arbug: Early return make the logic hard to read. We should refactor the code again in some time. But for now, please add a comment to describe the behavior.

Connor1996: I'm afraid of neglect error silently makes us hardly know whether level-merge/fallback working properly? maybe print some logs? @yiwu-arbug

yiwu-arbug: make sense. log per key is too noisy though. how about adding an error count and print at the end?

Connor1996: looking good. and please add a debug log for error message

yiwu-arbug: > looking good. and please add a debug log for error message

@wujy-cs mind address the comment? Thanks.

wujy-cs: @yiwu-arbug 
Ok.
I didn't record erorr for fallback since it's normal to read a GC-ed blob file and get error during fallback.

Connor1996:
Please fix the sign-off

wujy-cs:
> Please fix the sign-off

I signed, don't know why it's not work

Connor1996:
> > Please fix the sign-off
> 
> @Connor1996 I have signed, don't know why it doesn't work

You should sign all previous commits

yiwu-arbug:
Let me just force merge..

PR #120, opened 2019-11-27
Store Titan stats in rocksdb statistics

Connor1996:
Store Titan stats in rocksdb statistics, so we can reuse the implementation of Rocksdb StatisticsImpl


Connor1996:
blocked by  https://github.com/tikv/rocksdb/pull/136

yiwu-arbug:
CI still complain about `Assertion `tickerType < TICKER_MAX' failed.`. Can you check?

yiwu-arbug:
seems codecov adds coverage checks now? Maybe we should disable them.

PR #121, opened 2019-11-29
[PCP-21] Titan GC doesn’t affect online write

tabokie:
# Green Blob GC by merge operator

tikv/tikv#5739

* New value type: kMergeBlobIndex
* New value content: LSN + BlobIndex

## Merge Semantics: keep freshest version
**A. key + merged index**: source-index must be older than key, apply key

**B. index + merged index**: if index is stale (GCed), apply merged-index or simply delete key; else apply index

implementation:

* if any merge-index is sourced from base-index, it's certain that base-index is stale.
* `source-file-number` info can be stored in merge-blob-index

**C. merged index + merged index (merge against delete)**: deletes key (use deletion marker instead)

**D. merged index + merged index (partial merge)**: by LSN version of original *put*.

implementation:

* embed LSN in merge operand, during blob building append LSN to blob key, during GC pass this LSN to merge operand (also to new blob file).
* ~embed LSN in blob key, when conflict read blob for latest index. (partial merge must read blob)~

## GC workflow:

Three approaches to merge new index into rocksdb, requires further benchmarking:

* plain merge
* ingestion
* fast ingestion
  * write SST of newly created blob index as `Merge` op
  * ingest external SST in a weak consistent way (do not pause write and flush on overlap, only guarantee ingested file is latest in persistent level)

## Rocksdb Change:

tikv/rocksdb#147

* ~support kTypeBlobIndex in merge operation~ (to support user-specified merge on top of gc-merge, extend merge value to include operand type)
* add value type input/output to merge interface
* ~merge operator deletes key~
* skip_memtable_check flag in IngestOption

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,6 +93,7 @@ if (WITH_TITAN_TESTS AND (NOT CMAKE_BUILD_TYPE STREQUAL "Release"))
         blob_format_test
         blob_gc_job_test
         blob_gc_picker_test
+        gc_stats_test
         table_builder_test
         thread_safety_test
         titan_db_test
@@ -104,12 +105,12 @@ if (WITH_TITAN_TESTS AND (NOT CMAKE_BUILD_TYPE STREQUAL "Release"))
         rocksdb
         testutillib
         gtest)
-        
+
   foreach(test ${TESTS})
     add_executable(titan_${test} src/${test}.cc $<TARGET_OBJECTS:testharness>)
       target_link_libraries(titan_${test} ${TEST_LIBS})
       add_test(titan_${test} titan_${test})
-  endforeach(test ${TESTS})  
+  endforeach(test ${TESTS})
 endif()
 
 if (WITH_TITAN_TOOLS)
@@ -137,3 +138,16 @@ if (WITH_TITAN_TOOLS)
   target_include_directories(titandb_bench PRIVATE ${gflags_INCLUDE_DIR})
   target_link_libraries(titandb_bench ${TOOLS_LIBS})
 endif()
+
+# Installation - copy lib/ and include/
+
+include(GNUInstallDirs)
+install(DIRECTORY include/titan
+  COMPONENT devel
+  DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
+)
+install(TARGETS titan
+  COMPONENT devel
+  ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
+  INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
+)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,11 +5,11 @@ enable_language(C)
 find_package(Git)
 
 if (NOT ROCKSDB_GIT_REPO)
-  set(ROCKSDB_GIT_REPO "https://github.com/pingcap/rocksdb.git")
+  set(ROCKSDB_GIT_REPO "https://github.com/tabokie/rocksdb.git")
 endif()
 
 if (NOT ROCKSDB_GIT_BRANCH)
-  set(ROCKSDB_GIT_BRANCH "6.4.tikv")
+  set(ROCKSDB_GIT_BRANCH "green-blob-gc")
 endif()
 
 if (NOT DEFINED ROCKSDB_DIR)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,11 +5,11 @@ enable_language(C)
 find_package(Git)
 
 if (NOT ROCKSDB_GIT_REPO)
-  set(ROCKSDB_GIT_REPO "https://github.com/tabokie/rocksdb.git")
+  set(ROCKSDB_GIT_REPO "https://github.com/pingcap/rocksdb.git")
 endif()
 
 if (NOT ROCKSDB_GIT_BRANCH)
-  set(ROCKSDB_GIT_BRANCH "green-blob-gc")
+  set(ROCKSDB_GIT_BRANCH "6.4.tikv")
 endif()
 
 if (NOT DEFINED ROCKSDB_DIR)
@@ -93,6 +93,7 @@ if (WITH_TITAN_TESTS AND (NOT CMAKE_BUILD_TYPE STREQUAL "Release"))
         blob_format_test
         blob_gc_job_test
         blob_gc_picker_test
+        blob_index_merge_operator_test
         gc_stats_test
         table_builder_test
         thread_safety_test
--- a/src/blob_format.cc
+++ b/src/blob_format.cc
@@ -127,57 +127,39 @@ Status BlobIndex::DecodeFrom(Slice* src) {
   return s;
 }
 
-bool operator==(const BlobIndex& lhs, const BlobIndex& rhs) {
-  return (lhs.file_number == rhs.file_number &&
-          lhs.blob_handle == rhs.blob_handle);
+bool BlobIndex::operator==(const BlobIndex &rhs) const {
+  return (file_number == rhs.file_number && blob_handle == rhs.blob_handle);
 }
 
-void VersionedBlobIndex::EncodeTo(std::string* dst) const {
-  dst->push_back(kBlobRecord);
-  PutVarint64(dst, file_number);
-  blob_handle.EncodeTo(dst);
+void MergeBlobIndex::EncodeTo(std::string *dst) const {
+  BlobIndex::EncodeTo(dst);
   PutVarint64(dst, sequence);
+  PutVarint64(dst, source_file_number);
 }
 
-void VersionedBlobIndex::EncodeToUnversioned(std::string* dst) const {
-  dst->push_back(kBlobRecord);
-  PutVarint64(dst, file_number);
-  blob_handle.EncodeTo(dst);
+void MergeBlobIndex::EncodeToBase(std::string *dst) const {
+  BlobIndex::EncodeTo(dst);
 }
 
-Status VersionedBlobIndex::DecodeFrom(Slice* src) {
-  unsigned char type;
-  if (!GetChar(src, &type) || type != kBlobRecord ||
-      !GetVarint64(src, &file_number)) {
-    return Status::Corruption("VersionedBlobIndex");
-  }
-  Status s = blob_handle.DecodeFrom(src);
+Status MergeBlobIndex::DecodeFrom(Slice *src) {
+  Status s = BlobIndex::DecodeFrom(src);
   if (!s.ok()) {
-    return Status::Corruption("VersionedBlobIndex", s.ToString());
+    return s;
   }
-  if (!GetVarint64(src, &sequence)) {
-    return Status::Corruption("VersionedBlobIndex(sequence)");
+  if (!GetVarint64(src, &sequence) || !GetVarint64(src, &source_file_number)) {
+    return Status::Corruption("MergeBlobIndex");
   }
   return s;
 }
 
-Status VersionedBlobIndex::DecodeFromUnversioned(Slice* src) {
-  unsigned char type;
-  if (!GetChar(src, &type) || type != kBlobRecord ||
-      !GetVarint64(src, &file_number)) {
-    return Status::Corruption("VersionedBlobIndex");
-  }
-  Status s = blob_handle.DecodeFrom(src);
-  if (!s.ok()) {
-    return Status::Corruption("VersionedBlobIndex", s.ToString());
-  }
-  sequence = 0;
-  return s;
+Status MergeBlobIndex::DecodeFromBase(Slice *src) {
+  return BlobIndex::DecodeFrom(src);
 }
 
-bool operator==(const VersionedBlobIndex& lhs, const VersionedBlobIndex& rhs) {
-  return (lhs.file_number == rhs.file_number &&
-          lhs.blob_handle == rhs.blob_handle && lhs.sequence == rhs.sequence);
+bool MergeBlobIndex::operator==(const MergeBlobIndex &rhs) const {
+  return (sequence == rhs.sequence &&
+          source_file_number == rhs.source_file_number &&
+          BlobIndex::operator==(rhs));
 }
 
 void BlobFileMeta::EncodeTo(std::string* dst) const {
--- a/src/blob_format.cc
+++ b/src/blob_format.cc
@@ -284,12 +284,6 @@ void BlobFileMeta::FileStateTransit(const FileEvent& event) {
   }
 }
 
-void BlobFileMeta::AddDiscardableSize(uint64_t _discardable_size) {
-  assert(_discardable_size < file_size_);
-  discardable_size_ += _discardable_size;
-  assert(discardable_size_ < file_size_);
-}
-
 TitanInternalStats::StatsType BlobFileMeta::GetDiscardableRatioLevel() const {
   auto ratio = GetDiscardableRatio();
   TitanInternalStats::StatsType type;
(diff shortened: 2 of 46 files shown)


tabokie:
A bug is spotted in current work: merge's correctness depends on retrieving up-to-date logical file state, meanwhile the file state transit depends on all new indexes being inserted so we can use `LatestSequenceNumber` as obsolete sequence. The problem is tricky to solve given titan inclines to atomically change file state and persist it to log.
New idea is to avoid file state inspection (which also save us bunch of locking). Instead, embed old file number to merge operands, so that base index is ignored if sharing the same file number with merges (announcing itself the source index producing this merge operand).
**New value content is now LSN + old file number + BlobIndex**

tabokie:
ready for review now @yiwu-arbug @Connor1996. benchmark comming soon.

yiwu-arbug:
@tabokie Awesome! Let's see how benchmark is doing.

tabokie:
@yiwu-arbug @Connor1996 Here are some test results, Write performance brought by new gc is impressive, while gc-with-merge is more feasible with read regression below 10 percent.

**dbbench** (`./titandb_bench --key_size="16" --value_size="256" --cache_size="4294967296" --titan_min_blob_size="0" --num="10000000" --numdistinct="10000" --benchmarks="fillseq,updaterandom,readrandom" --threads="6" --titan_max_background_gc="6"`):
gc_rewrite_mode = kDefault
fillseq: 99.3 MB/s; overwrite: 23.2 MB/s; readrandom: 130.5 MB/s
gc_rewrite_mode = kMerge
fillseq: 102.5 MB/s; overwrite: 59.3 MB/s; readrandom: 122.8 MB/s
gc_rewrite_mode = kFastIngest
fillseq: 101.1 MB/s; overwrite: 61.8 MB/s; readrandom: 82.4 MB/s

**go-ycsb** (`max-background-gc=4, threads=256, readcount=1kw, operationcount=2kw`)
gc_rewrite_mode = kDefault
update ops 5w -> 4.53w, read ops 12w
gc_rewrite_mode = kMerge
update ops 5w -> 4.58w, read ops 11w
gc_rewrite_mode = kFastIngest
update ops 52 -> 4.7w, read ops 7w

CMakeLists.txt
```diff
@@ -5,11 +5,11 @@ enable_language(C)
 find_package(Git)
 
 if (NOT ROCKSDB_GIT_REPO)
-  set(ROCKSDB_GIT_REPO "https://github.com/pingcap/rocksdb.git")
+  set(ROCKSDB_GIT_REPO "https://github.com/tabokie/rocksdb.git")
```
Connor1996: any patch not included in our branch?

src/blob_gc_job.cc
```diff
@@ -433,32 +483,83 @@ Status BlobGCJob::RewriteValidKeyToLSM() {
...
-      metrics_.gc_bytes_relocated += write_batch.second.blob_record_size();
-      // Key is successfully written to LSM.
-    } else if (s.IsBusy()) {
-      metrics_.gc_num_keys_overwritten++;
-      metrics_.gc_bytes_overwritten += write_batch.second.blob_record_size();
-      // The key is overwritten in the meanwhile. Drop the blob record.
-    } else {
-      // We hit an error.
+    case TitanGcRewriteMode::kIngest:
+    case TitanGcRewriteMode::kFastIngest:
```
Connor1996: check dropped column family and shutting down?

src/blob_gc_job.cc
```diff
@@ -433,32 +483,83 @@ Status BlobGCJob::RewriteValidKeyToLSM() {
...
+        }
+        s = db_impl->Write(wo, &write_batch);
+        if (s.ok()) {
+          // count written bytes for new blob index.
+          metrics_.bytes_written += write_batch.GetDataSize();
+          metrics_.gc_num_keys_relocated++;
+          // metrics_.gc_bytes_relocated +=
+          // write_batch.second.blob_record_size();
+          // Key is successfully written to LSM.
+        } else if (s.IsBusy()) {
```
Connor1996: seems `IsBusy` doesn't mean `overwritten in the meanwhile` for writing with merge ops.

src/blob_gc_job.cc
```diff
@@ -222,31 +239,64 @@ Status BlobGCJob::DoRunGC() {
...
+    }
     if (!s.ok()) {
       break;
     }
   }
 
+  if (s.ok() && gc_rewrite_mode_ >= TitanGcRewriteMode::kIngest) {
+    if (entry_count > 0) {
+      ingestion_file_status_ = blob_index_table.Finish();
+      if (s.ok()) {
```
Connor1996: shoud check `ingestion_file_status_` ?

src/blob_gc_job.cc
```diff
@@ -222,31 +239,64 @@ Status BlobGCJob::DoRunGC() {
...
+      }
+    }
     if (!s.ok()) {
       break;
     }
   }
 
+  if (s.ok() && gc_rewrite_mode_ >= TitanGcRewriteMode::kIngest) {
+    if (entry_count > 0) {
+      ingestion_file_status_ = blob_index_table.Finish();
```
Connor1996: how about calling it in `RewriteValidKeyToLSM`,  so we can remove `ingestion_file_ready_`

src/blob_format.cc
```diff
@@ -19,19 +19,21 @@ bool GetChar(Slice* src, unsigned char* value) {
 
 void BlobRecord::EncodeTo(std::string* dst) const {
   PutLengthPrefixedSlice(dst, key);
+  PutVarint64(dst, sequence);
```
Connor1996: Please consider compatibility issue

src/blob_format.h
```diff
@@ -139,12 +141,31 @@ struct BlobIndex {
...
   void EncodeTo(std::string* dst) const;
+  void EncodeToBase(std::string* dst) const;
   Status DecodeFrom(Slice* src);
+  Status DecodeFromBase(Slice* src);
 
-  friend bool operator==(const BlobIndex& lhs, const BlobIndex& rhs);
+  bool operator==(const MergeBlobIndex& rhs) const;
 };
 
+// NO.merge
```
Connor1996: what's the meaning of the comment?

src/db_impl.h
```diff
@@ -268,6 +268,7 @@ class TitanDBImpl : public TitanDB {
   EnvOptions env_options_;
   DBImpl* db_impl_;
   TitanDBOptions db_options_;
+  TitanGcRewriteMode gc_rewrite_mode_{TitanGcRewriteMode::kMerge};
```
Connor1996: let's not enable it by default at first

src/blob_index_merge_operator.h
```diff
@@ -0,0 +1,95 @@
...
+#include "blob_file_set.h"
+
+namespace rocksdb {
+namespace titandb {
+
+class BlobIndexMergeOperator : public MergeOperator {
+ public:
+  BlobIndexMergeOperator() = default;
+
+  bool FullMergeV2(const MergeOperationInput& merge_in,
```
Connor1996: could you please add some comments or diagram to illustrate the cases for merge operator?

src/db_impl.cc
```diff
@@ -387,6 +389,7 @@ Status TitanDBImpl::CreateColumnFamilies(
   std::vector<ColumnFamilyDescriptor> base_descs;
   std::vector<std::shared_ptr<TableFactory>> base_table_factory;
   std::vector<std::shared_ptr<TitanTableFactory>> titan_table_factory;
+  std::vector<std::shared_ptr<BlobIndexMergeOperator>> merge_operators;
```
yiwu-arbug: Seem to me you can set the same merge operator object to all CFs. We can keep a shared_ptr at TitanDBImpl for reuse.

yiwu-arbug:
@tabokie As we are approaching the end of PCP, we need to merge this PR this week. To speed it up, let's do the following:
1. remove all ingest file related logic to make review simpler.
2. send PR for related rocksdb changes to tikv/rocksdb repo.
3. Add option in options.h to turn on/off merge op GC, and default to off.
3. adding sequence to blob record is a breaking change. can you remind us why it is needed? I'm incline to remove it for now, even if it will introduce correctness issue.
4. can you remind us on the merge semantics? I find the following in the old docs but it is still unclear to me:
Screen Shot 2020-02-09 at 1 15 14 PM

This is great work overall! We can keep working on remaining work after PCP. We definitely want it to included in tikv 4.0 release.

src/blob_format.h
tabokie: filename format, not needed now.

src/blob_index_merge_operator.h
```diff
@@ -0,0 +1,95 @@
...
+                   MergeOperationOutput* merge_out) const override {
+    Status s;
+    if (merge_in.existing_value && merge_in.value_type == kTypeValue) {
```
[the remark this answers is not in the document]
tabokie: Yes, and this input is neccessary for our use case.

src/db_impl.cc
tabokie: Indeed, now the operator is completely stateless.

tabokie:
@yiwu-arbug Some Q&A first:
1. Embedding LSN in blob record is a critical design choice. In this [snippet](https://github.com/tikv/titan/blob/bb813707ad10d5ce3599faf4912ffec7500a74de/src/blob_index_merge_operator.h#L64-L70), merge operator avoids losing blob index by only keeping the latest one in multiple merge operands. Here, *latest* refer to timestamp of the original *Put*, which is a meaningful concept only on titan end, and therefore required to be stored in blob record payload. In detail, this `blob timestamp` is flowing in this path: `MemKey ->[table-builder/GC-by-put]-> BlobRecord ->[GC-by-merge/merge]->MergeOperand`. Notice the `GC-by-put` mutation is causing mis-alignment between the two timestamp.
2. On a second thought, merge semantics is better explained by describing the anomalies it tries to avoid (`[X]` refers to *put*, `(X)` refers to *merge*, full content is `[X blob-ts source-fno value-fno]`, time is flowing rightwards):
*. basic rule, keep the base value: `[?] ... [X] (Y) (Z)` => `[X]`
this is mostly true when `Y/Z` stem from older `?`, guaranteed by rocksdb visibility constraint.
a. base + delta: `[X] (Y) (X')` => `[X']`
simple equal case where `X'` stem from exactly `X`. We identify this case by checking the source-file-number stored in merge operands, in detail: `[X - - 0] (Y) (X' - 0 1)`.
b. merge reorder and gc merge operand: `[A] [B] (B') (A)` => `[B"]`
visibility order of merge != that of value. We need to preserve original order in reordered merge operands (explained before). in detail: `[A 0 - -] [B 1 - -] (B' 1 - -) (A' 0 - -)`
c. deletion: `[X] (X) [delete]` => `[delete marker]`
introducing deletion marker is a workaround against a restriction that rocksdb merge can't produce empty result.

tabokie:
Almost good now, PTAL @yiwu-arbug @Connor1996

yiwu-arbug:
Haven't go through the rocksdb side of change, but it seems to be a large one. Are we able to get a state that we don't depend on rocksdb change, but have minimum correctness issue?

src/db_iter.h
```diff
@@ -132,7 +132,30 @@ class TitanDBIterator : public Iterator {
                       status_.ToString().c_str());
       return;
     }
+    while (BlobIndex::IsDeletionMarker(index)) {
+      // skip deletion marker
+      if (forward) {
+        iter_->Next();
+      } else {
+        iter_->Prev();
+      }
```
yiwu-arbug: update `BLOB_DB_NUM_NEXT` and `BLOB_DB_NUM_PREV`

src/blob_index_merge_operator.h
```diff
@@ -0,0 +1,106 @@
...
+        existing_index_valid = false;
+      }
+      // if base index is still valid, merges must be sourced from older
+      // modifies.
+      if (!existing_index_valid) {
+        if (!filled) {
+          merge_index = index;
+          latest_sequence = index.sequence;
+          filled = true;
+        } else if (latest_sequence <= index.sequence) {
```
yiwu-arbug: Can you give an example of `latest_sequence < index.sequence` happens? It seems to me merges will never come with a sequence > existing_value_seq.

yiwu-arbug:
I may be missing something, but I still think merge operator don't need to know LSN of values. Just like the write callback for gc-with-put doesn't need to know LSN of keys. I think we can simply use the following logic in merge op (pseudo-code):
```
blob_index = existing_value;
for (merge_op : in_merge_op_list) {
    if (merge_op.source_file == blob_index.file && merge_op.source_offset == blob_index.offset) {
        blob_index.file = merge_op.file;
        blob_index.offset = merge_op.offset
    }
}
```

src/blob_index_merge_operator.h
tabokie: Make sense

src/blob_index_merge_operator.h
```diff
@@ -0,0 +1,106 @@
...
+        }
+      }
+    }
+    merge_out->new_type = kTypeBlobIndex;
+    if (existing_index_valid) {
+      merge_out->existing_operand = *merge_in.existing_value;
+    } else {
+      assert(filled);
+      merge_out->new_value.clear();
+      merge_index.EncodeTo(&merge_out->new_value);
```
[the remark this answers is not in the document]
tabokie: `merge_index` is a normal BlobIndex type.

src/blob_index_merge_operator.h
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    MergeBlobIndex index;
+    BlobIndex merge_index;
+    for (auto operand : merge_in.operand_list) {
+      s = index.DecodeFrom(&operand);
+      if (!s.ok()) {
+        return false;
+      }
+      if (existing_index_valid) {
```
Connor1996: can write like that
```C++
   for (auto operand : merge_in.operand_list) {
      s = index.DecodeFrom(&operand);
      if (!s.ok()) {
        return false;
      }
      if (index.source_file_number == existing_index.file_number &&
          index.source_file_offset == existing_index.blob_handle.offset) {
        existing_index_valid = false;
        merge_index = index;
      }
    }
    merge_out->new_type = kBlobIndex;
    if (existing_index_valid) {
      merge_out->existing_operand = *merge_in.existing_value;
    } else {
      merge_out->new_value.clear();
      merge_index.EncodeTo(&merge_out->new_value);
    }

```

src/blob_index_merge_operator.h
```diff
@@ -0,0 +1,93 @@
...
+  // and *src-file-offset*.
+  // Merge operation follows such rules:
+  // *. basic rule (keep base value): [Y][Z] ... [X](Y)(Z) => [X]
+  // a. same put (keep merge value): [Y] ... [X](Y)(X')(X") => [X"]
+  //    we identify this case by checking *src-location* of merges against
+  //    *blob-handle* of base.
+  // b. deletion (keep deletion marker): [delete](X)(Y) => [deletion marker]
+  //    this is a workaround since vanilla rocksdb disallow empty result from
+  //    merge.
+  bool FullMergeV2(const MergeOperationInput& merge_in,
```
Connor1996: better to have a test for merge operator

src/blob_index_merge_operator.h
tabokie: Notice the second hand uses `index.source_file_number == merge_index.file_number`, the *derive-from* is chained from base to multiple operands.

yiwu-arbug:
LGTM. Leave to @Connor1996 for merge.

zhangjinpeng1987:
@Connor1996 Can we merge it now?

PR #123, opened 2019-12-01
fix name shadowing issue

tabokie:
fix compiling error: 
> error: declaration of ‘initialized’ shadows a member of 'this'

Signed-off-by: tabokie <xy.tao@outlook.com>

PR #125, opened 2019-12-03
fix compile error caused by name

Little-Wallace:
changed param name to pass CI.

Signed-off-by: Little-Wallace <bupt2013211450@gmail.com>


yiwu-arbug:
Fixed in #123

PR #130, opened 2019-12-21
Reconstruct GC stats on reopen

yiwu-arbug:
Summary:
* On DB reopen, iterate all SST file's properties and sum up live data size of all blob files to use as GC stats.
* Remove previous logic to set GC mark on blob files on reopen.
* Refactor `OnFlushCompleted` and `OnCompactionCompleted`. Check file state is `kPendingLSM` before transit file state. 
* Refactor `BlobFileMeta`.

Test Plan:
Updated unit tests

### DIFF
--- a/src/db_impl.cc
+++ b/src/db_impl.cc
@@ -743,8 +743,8 @@ Status TitanDBImpl::DeleteFilesInRanges(ColumnFamilyHandle* column_family,
     // Get all the files within range except L0, cause `DeleteFilesInRanges`
     // would not delete the files in L0.
     for (int level = 1; level < vstorage->num_non_empty_levels(); level++) {
-      if (vstorage->LevelFiles(i).empty() ||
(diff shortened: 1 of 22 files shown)


yiwu-arbug:
ready for review. though I'm still working on adding tests.

src/db_impl.cc
```diff
@@ -1187,87 +1165,92 @@ void TitanDBImpl::OnCompactionCompleted(
...
+                         cf_options.blob_file_discardable_ratio) {
+            file->FileStateTransit(BlobFileMeta::FileEvent::kNeedMerge);
+          } else {
+            if (count_sorted_run) {
+              to_merge_candidates.push_back(file);
+            }
+          }
         }
       }
+      if (file->file_state() == BlobFileMeta::FileState::kNormal) {
```
Connor1996: why not include `kToMerge`?

yiwu-arbug: Fixed

PR #135, opened 2020-01-22
Fix GC stat not being updated after DeleteFilesInRange

yiwu-arbug:
Summary:
Fix a typo in `TitanDBImpl::DeleteFilesInRange()` which makes GC stats not being updated after `DeleteFilesInRange()` call.

Test Plan:
See the new test

Signed-off-by: Yi Wu <yiwu@pingcap.com>

PR #137, opened 2020-01-30
Fix GC stat not being updated after DeleteFilesInRange (#135)

yiwu-arbug:
Summary:
Fix a typo in `TitanDBImpl::DeleteFilesInRange()` which makes GC stats not being updated after `DeleteFilesInRange()` call.

Test Plan:
See the new test


yiwu-arbug:
cherry-picking #135 to tikv-3.x branch. We'll fast-forward tikv-4.x branch later so no need to cherry-pick there.

PR #138, opened 2020-01-31
Fix build on Windows

kennytm:
1. use `#include "port/port.h"` instead of `#include "port/port_posix.h"`
2. added an "install" target for CMake to workaround alexcrichton/cmake-rs#18
3. MSVC reported 2 warnings in Titan (and tons of warnings in RocksDB), and these are fixed.


yiwu-arbug:
@kennytm do you only need this for tikv-4.x branch or also need it for tikv-3.x?

kennytm:
@yiwu-arbug Not required for 3.x for now, but I don't see any harm cherry-picking. Just trying to see how far can we build TiKV on Windows using MSVC without WSL.

Issue #127, opened 2019-12-09
Implement PrioritizedCache


yiwu-arbug:
Implement `PrioritizedCache`, which wraps a single `LRUCache`, but have additional API to return child `Cache` instances that:
* HighPriCache: always insert into the cache with high-priority, regardless of user provided option
* LowPriCache: always insert into the cache with low-priority, regardless of user provided option

Part of https://github.com/tikv/tikv/issues/5742

yiwu-arbug:
cc @haoxiang47

haoxiang47:
Hi @yiwu-arbug , sorry for late for this PCP, I had simplily add a PR: https://github.com/tikv/rocksdb/pull/144 for this issue, add HighPriCache and LowPriCache two interface in rocksdb

yiwu-arbug:
> Hi @yiwu-arbug , sorry for late for this PCP, I had simplily add a PR: [tikv/rocksdb#144](https://github.com/tikv/rocksdb/pull/144) for this issue, add HighPriCache and LowPriCache two interface in rocksdb

Thank much for the work! Commented on the PR.

Connor1996:
As blob specific cache priority is introduced, close it https://github.com/tikv/rocksdb/pull/354

PR #142, opened 2020-02-21
add prioritized cache wrapper

haoxiang47:
add a prioritized cache wrapper, for this issue: https://github.com/tikv/titan/issues/127

### DIFF
--- a/include/titan/prioritized_cache.h
+++ b/include/titan/prioritized_cache.h
@@ -1,0 +1,117 @@
+#pragma once
+#include "rocksdb/cache.h"
+
+namespace rocksdb {
+namespace titandb {
+
+class PrioritizedCache : public Cache {
+ public:
+  // Constructs a PrioritizedCache, which is a wrapper
+  // of rocksdb Cache
+  PrioritizedCache(std::shared_ptr<Cache> cache, Cache::Priority priority);
+  ~PrioritizedCache();
+
+  // Get the Cache ptr
+  std::shared_ptr<Cache> GetCache();
+
+  // always insert into the cache with the priority when init,
+  // regardless of user provided option
+  Status Insert(const Slice& key, void* value, size_t charge,
+                void (*deleter)(const Slice& key, void* value),
+                Cache::Handle** handle, Cache::Priority /*priority*/) override;
+
+  // The type of the Cache
+  const char* Name() const override;
+
+  // If the cache has no mapping for "key", returns nullptr.
+  //
+  // Else return a handle that corresponds to the mapping.  The caller
+  // must call this->Release(handle) when the returned mapping is no
+  // longer needed.
+  // If stats is not nullptr, relative tickers could be used inside the
+  // function.
+  Cache::Handle* Lookup(const Slice& key, Statistics* stats = nullptr) override;
+
+  // Increments the reference count for the handle if it refers to an entry in
+  // the cache. Returns true if refcount was incremented; otherwise, returns
+  // false.
+  // REQUIRES: handle must have been returned by a method on *this.
+  bool Ref(Cache::Handle* handle) override;
+
+  /**
+   * Release a mapping returned by a previous Lookup(). A released entry might
+   * still  remain in cache in case it is later looked up by others. If
+   * force_erase is set then it also erase it from the cache if there is no
+   * other reference to  it. Erasing it should call the deleter function that
+   * was provided when the
+   * entry was inserted.
+   *
+   * Returns true if the entry was also erased.
+   */
+  // REQUIRES: handle must not have been released yet.
+  // REQUIRES: handle must have been returned by a method on *this.
+  bool Release(Cache::Handle* handle, bool force_erase = false) override;
+
+  // Return the value encapsulated in a handle returned by a
+  // successful Lookup().
+  // REQUIRES: handle must not have been released yet.
+  // REQUIRES: handle must have been returned by a method on *this.
+  void* Value(Cache::Handle* handle) override;
+
+  // If the cache contains entry for key, erase it.  Note that the
+  // underlying entry will be kept around until all existing handles
+  // to it have been released.
+  void Erase(const Slice& key) override;
+
+  // Return a new numeric id.  May be used by multiple clients who are
+  // sharding the same cache to partition the key space.  Typically the
+  // client will allocate a new id at startup and prepend the id to
+  // its cache keys.
+  uint64_t NewId() override;
+
+  // sets the maximum configured capacity of the cache. When the new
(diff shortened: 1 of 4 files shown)


haoxiang47:
@yiwu-arbug sorry for  a little late, I have update this PR to add a wrapper called PrioritizedCache, has two function `HighPriCacheInsert` and `LowPriCacheInsert` which can insert high and low cache priorized regardless of user provided option, and a interface `GetCache` to return the `Cache` ptr

src/prioritized_cache.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+#include "rocksdb/cache.h"
+#include "rocksdb/env.h"
+
+namespace rocksdb {
+namespace titandb {
+
+class PrioritizedCache {
```
yiwu-arbug: Sorry I didn't explain clearly again. What I want is something like this. Do you mind update the code? Thanks.
```
class PrioritizedCache: public Cache {
   public:
      PrioritizedCache(std::shared_ptr<Cache> cache, Cache::Priority priority)
          : cache_(cache), priority_(priority) {}

      Status Insert(..., Cache::Priority /*priority*/) override {
        return Insert(..., priority_);
      }

   private:
      std::shared_ptr<Cache> cache_;
      Cache::Priority priority_;
};
```

yiwu-arbug: With it  we can use the same underlying LRUCache for both rocksdb block_cache and titan blob_cache, but forcing block_cache to always insert with high-pri, and titan always insert with low-pri. This is because the data in rocksdb is the index for titan values, so they are more valuable and we want them to in cache more.

haoxiang47: Oh, I got it, I will update the code soon.

haoxiang47:
@yiwu-arbug I have updated the code, please review again

include/titan/prioritized_cache.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+#include "rocksdb/cache.h"
+
+namespace rocksdb {
+namespace titandb {
+
+class PrioritizedCache : Cache {
```
yiwu-arbug: s/`Cache`/`public Cache`/

and implement all other methods in `Cache` interface, by passing arguments directly to underlying `cache_`.

src/prioritized_cache.cc
```diff
@@ -0,0 +1,23 @@
...
+
+PrioritizedCache::~PrioritizedCache() {}
+
+std::shared_ptr<Cache> PrioritizedCache::GetCache() { return cache_; }
+
+Status PrioritizedCache::Insert(const Slice& key, void* value, size_t charge,
+                                void (*deleter)(const Slice& key, void* value),
+                                Cache::Handle** handle,
+                                Cache::Priority /*priority*/) {
+  if (!cache_) return Status::NotFound();
```
yiwu-arbug: let's assume `cache_` always present.

haoxiang47: ok, so I add an assert(cache_) here to replace check?

include/titan/prioritized_cache.h
haoxiang47: ok, I have updated again now, please review again @yiwu-arbug

yiwu-arbug:
Mind adding some very simple unit test? Like have a mock underlying Cache object that checks what's the priority pass to it on Insert.

include/titan/prioritized_cache.h
yiwu-arbug: I think you need to implement all methods (`Lookup`, `Release`, `Erase`, etc) in the `Cache` interface.

src/prioritized_cache_test.cc
```diff
@@ -0,0 +1,61 @@
...
+  ASSERT_TRUE(lo_ok.ok());
+
+  // here we set insert Priority to low, to check if it will 
+  // insert to high in fact
+  auto hi_ok = high_cache.Insert(high_cache_key, (void*)&cacahe_value, cache->GetCapacity(), 
+      &DeleteCacheValue<Slice>, &cache_handle, Cache::Priority::LOW);
+  ASSERT_TRUE(hi_ok.ok());
+
+  auto high_handle = cache->Lookup(high_cache_key);
+  if (high_handle) {
```
yiwu-arbug: Assert high_handle != nullptr.

src/prioritized_cache_test.cc
```diff
@@ -0,0 +1,61 @@
...
+    auto v = reinterpret_cast<Slice*>(cache->Value(high_handle));
+    ASSERT_EQ(cacahe_value.data(), v->data());
+  }
+
+  auto low_handle = cache->Lookup(low_cache_key);
```
yiwu-arbug: assert low_handle == nulllptr.

src/prioritized_cache_test.cc
```diff
@@ -0,0 +1,61 @@
...
+      &DeleteCacheValue<Slice>, &cache_handle, Cache::Priority::LOW);
+  ASSERT_TRUE(lo_ok.ok());
+
+  // here we set insert Priority to low, to check if it will 
+  // insert to high in fact
+  auto hi_ok = high_cache.Insert(high_cache_key, (void*)&cacahe_value, cache->GetCapacity(), 
+      &DeleteCacheValue<Slice>, &cache_handle, Cache::Priority::LOW);
+  ASSERT_TRUE(hi_ok.ok());
+
+  auto high_handle = cache->Lookup(high_cache_key);
```
yiwu-arbug: lookup through `high_cache` instead of `cache`.

src/prioritized_cache_test.cc
yiwu-arbug: lookup through `low_cache` instead of `cache`.

src/prioritized_cache_test.cc
```diff
@@ -0,0 +1,61 @@
...
+  if (high_handle) {
+    auto v = reinterpret_cast<Slice*>(cache->Value(high_handle));
+    ASSERT_EQ(cacahe_value.data(), v->data());
```
yiwu-arbug: Should release the handle here, otherwise there's memory leak: `high_cache->Release(high_handle)`. This should fix CI.

include/titan/prioritized_cache.h
```diff
@@ -0,0 +1,117 @@
+#pragma once
```
yiwu-arbug: Please add the header to new files:
```
// Copyright 2020 TiKV Project Authors. Licensed under Apache-2.0.
```

Connor1996:
As blob-specific cache priority is introduced, close it https://github.com/tikv/rocksdb/pull/354

PR #151, opened 2020-03-25
Add cf abtest for Titan

Connor1996:
For cf consistency test, make some column families enable Titan and others not.

### DIFF
--- a/tools/titandb_stress.cc
+++ b/tools/titandb_stress.cc
@@ -2942,14 +2942,13 @@ class StressTest {
         options_.max_background_gc = FLAGS_max_background_gc;
         titandb::TitanDBOptions titan_db_options(options_);
         std::vector<titandb::TitanCFDescriptor> titan_cf_descriptors;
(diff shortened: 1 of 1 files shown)


tools/titandb_stress.cc
```diff
@@ -2941,9 +2942,22 @@ class StressTest {
...
         titandb::TitanDBOptions titan_db_options(options_);
         std::vector<titandb::TitanCFDescriptor> titan_cf_descriptors;
+        Random rand(static_cast<uint32_t>(FLAGS_seed));
         for (const auto& cfd : cf_descriptors) {
           titan_cf_descriptors.emplace_back(
               titandb::TitanCFDescriptor{cfd.name, options_});
+          if (FLAGS_test_cf_consistency) {
+            // if test cf consistency, make some column families not store value
+            // in blob
+            if (rand.OneIn(2)) {
```
yiwu-arbug: instead of random, can we make it more deterministic? With a random value, it is possible all CF gets the same setting with bad luck.

PR #154, opened 2020-04-28
exclude metadata size from file size for discardable ratio

Connor1996:
exclude metadata size from file size, otherwise discardable ratio would be wrong when the blob file is small.


yiwu-arbug:
Can you take a look at travis failure?

PR #157, opened 2020-05-06
Fix discardable ratio count stat

Connor1996:
After creating a new blob file meta, the live data size is 0. The stat is counted when adding the blob file but with live_data_size == 0. So we should update the stat when the live data size is initialized.

### DIFF
--- a/src/db_impl.cc
+++ b/src/db_impl.cc
@@ -1233,7 +1233,9 @@ void TitanDBImpl::OnCompactionCompleted(
           assert(false);
           delta = 0;
         }
+        SubStats(stats_.get(), compaction_job_info.cf_id, file->GetDiscardableRatioLevel(), 1);
         file->set_live_data_size(static_cast<uint64_t>(delta));
+        AddStats(stats_.get(), compaction_job_info.cf_id, file->GetDiscardableRatioLevel(), 1);
         file->FileStateTransit(BlobFileMeta::FileEvent::kCompactionCompleted);
         to_merge_candidates.push_back(file);
         ROCKS_LOG_INFO(
--- a/src/db_impl.cc
+++ b/src/db_impl.cc
@@ -1233,9 +1233,11 @@ void TitanDBImpl::OnCompactionCompleted(
           assert(false);
           delta = 0;
         }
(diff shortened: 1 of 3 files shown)


src/db_impl.cc
```diff
@@ -1137,7 +1137,14 @@ void TitanDBImpl::OnFlushCompleted(const FlushJobInfo& flush_job_info) {
         assert(false);
         delta = 0;
       }
+
+      // the metrics is counted in the table builder when flushing,
+      // so update it when updateing the live data size.
+      SubStats(stats_.get(), flush_job_info.cf_id,
```
yiwu-arbug, line 1143: This is when we initially set `live_data_size` for the file. Where do we add stats before? I think we should not add stats before we do here.

src/db_impl_gc.cc
```diff
@@ -83,7 +83,11 @@ Status TitanDBImpl::InitializeGC(
           blob_storage->FindFile(file_size.first).lock();
       if (file != nullptr) {
         assert(file->live_data_size() == 0);
+        SubStats(stats_.get(), cf_handle->GetID(),
+                 file->GetDiscardableRatioLevel(), 1);
```
yiwu-arbug, line 87: ditto

src/db_impl.cc
Connor1996: https://github.com/tikv/titan/blob/927ee60912d7b39085849899e5e67d765c8bef35/src/blob_storage.cc#L89-L98
When applying the edit, I think it's the right place to update the stat which is more aligned to sematic logic.
So I just handle the special cases instead of removing the code where we do before.

src/db_impl.cc
```diff
@@ -1137,7 +1137,14 @@ void TitanDBImpl::OnFlushCompleted(const FlushJobInfo& flush_job_info) {
...
+               file->GetDiscardableRatioLevel(), 1);
       file->set_live_data_size(static_cast<uint64_t>(delta));
+      AddStats(stats_.get(), flush_job_info.cf_id,
```
yiwu-arbug, line 1146: should have the same change for OnCompactionCompleted.

PR #161, opened 2020-05-14
Fix wrong live data size when encounter rewrite failure

Connor1996:
Consider the case when GC is rewriting the new blob index to LSM, there is a concurrent front ground update the value of the blob index corresponding key. So the rewrite callback will fail, and the blob record is dropped silently.

The record is actually discarded, but we didn't update the live data size. So it may avoid the file being GCed.

### DIFF
--- a/src/blob_gc_job.cc
+++ b/src/blob_gc_job.cc
@@ -534,8 +534,6 @@ Status BlobGCJob::RewriteValidKeyToLSM() {
       }
       SubStats(stats_, cf_id, file->GetDiscardableRatioLevel(), 1);
       file->UpdateLiveDataSize(-blob_file.second);
-      fprintf(stderr, "update file %lu dropped %lu %lf", blob_file.first,
-              blob_file.second, file->GetDiscardableRatio());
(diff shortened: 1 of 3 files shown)


src/blob_gc_job.cc
```diff
@@ -508,6 +520,32 @@ Status BlobGCJob::RewriteValidKeyToLSM() {
...
+      auto file = blob_storage->FindFile(blob_file.first).lock();
+      if (!file) {
+        ROCKS_LOG_ERROR(db_options_.info_log,
+                        "Blob File %" PRIu64 " not found when GC.",
+                        blob_file.first);
+        continue;
+      }
+      SubStats(stats_, cf_id, file->GetDiscardableRatioLevel(), 1);
+      file->UpdateLiveDataSize(-blob_file.second);
+      fprintf(stderr, "update file %lu dropped %lu %lf", blob_file.first,
```
cosven, line 537: doubt why not use `ROCKS_LOG_BUFFER` macro here?

Connor1996: just forget to clean the log used for debug

PR #164, opened 2020-05-22
feat: Titan should return user value to compaction filter #163

gotoxu:
Signed-off-by: Xu Qiaolun <Jamesxql@Gmail.com>

### DIFF
--- a/src/compaction_filter.h
+++ b/src/compaction_filter.h
@@ -1,0 +1,111 @@
+#pragma once
+
+#include "db_impl.h"
+#include "rocksdb/compaction_filter.h"
+
+#include <utility>
+
+namespace rocksdb {
+namespace titandb {
+
+class TitanCompactionFilter final : public CompactionFilter {
+public:
+  explicit TitanCompactionFilter(const TitanDBImpl *db,
+                                 const CompactionFilter *origin, uint32_t cf_id)
+      : titan_db_impl_(db), origin_filter_(origin), column_family_id_(cf_id) {}
+
+  const char *Name() const override { return "TitanCompactionFilter"; }
+
+  Decision FilterV2(int level, const Slice &key, ValueType value_type,
+                    const Slice &value, std::string *new_value,
+                    std::string *skip_until) const override {
+    if (value_type != kBlobIndex) {
+      return origin_filter_->FilterV2(level, key, value_type, value, new_value,
+                                      skip_until);
+    }
+
+    BlobIndex blob_index;
+    Slice origin_value(value.data());
+    Status s = blob_index.DecodeFrom(&origin_value);
+    if (!s.ok()) {
+      // Unable to decode blob index. Keeping the value.
+      return Decision::kKeep;
+    }
+    if (BlobIndex::IsDeletionMarker(blob_index)) {
+      return Decision::kRemove;
+    }
+
+    BlobRecord record;
+    PinnableSlice buffer;
+
+    titan_db_impl_->mutex_.Lock();
+    auto storage =
+        titan_db_impl_->blob_file_set_->GetBlobStorage(column_family_id_)
+            .lock();
+    titan_db_impl_->mutex_.Unlock();
+
+    if (storage) {
+      ReadOptions read_options;
+      s = storage->Get(read_options, blob_index, &record, &buffer);
+    } else {
+      // Column family not found, remove the value.
+      return Decision::kRemove;
+    }
+
+    if (s.ok()) {
+      bool value_changed = false;
+      bool rv = origin_filter_->Filter(level, key, record.value, new_value,
+                                       &value_changed);
+      if (rv) {
+        return Decision::kRemove;
+      }
+      return value_changed ? Decision::kChangeValue : Decision::kKeep;
+    }
+
+    // GetBlobRecord failed, keep the value.
+    return Decision::kKeep;
+  }
+
+private:
+  const TitanDBImpl *titan_db_impl_;
+  const CompactionFilter *origin_filter_;
+  uint32_t column_family_id_;
+};
+
+class TitanCompactionFilterFactory final : public CompactionFilterFactory {
+public:
+  explicit TitanCompactionFilterFactory(const TitanDBImpl *db)
+      : titan_db_impl_(db), origin_filter_(nullptr),
+        origin_cf_factory_(nullptr) {}
+
+  const char *Name() const override { return "TitanCompactionFilterFactory"; }
+
+  void SetCF(const CompactionFilter *cf) { origin_filter_ = cf; }
+
+  void SetCFFactory(std::shared_ptr<CompactionFilterFactory> cf_factory) {
+    origin_cf_factory_ = std::move(cf_factory);
+  }
+
+  std::unique_ptr<CompactionFilter>
+  CreateCompactionFilter(const CompactionFilter::Context &context) override {
+    const CompactionFilter *compaction_filter = origin_filter_;
+    std::unique_ptr<CompactionFilter> compaction_filter_from_factory = nullptr;
+    if (compaction_filter == nullptr) {
+      // One of the origin_filter_ and origin_cf_factory_ must not be null
+      compaction_filter_from_factory =
+          origin_cf_factory_->CreateCompactionFilter(context);
+      compaction_filter = compaction_filter_from_factory.get();
+    }
+
+    return std::unique_ptr<CompactionFilter>(new TitanCompactionFilter(
+        titan_db_impl_, compaction_filter, context.column_family_id));
+  }
+
+private:
+  const TitanDBImpl *titan_db_impl_;
+  const CompactionFilter *origin_filter_;
+  std::shared_ptr<CompactionFilterFactory> origin_cf_factory_;
+};
+
+} // namespace titandb
+} // namespace rocksdb
--- a/src/compaction_filter.h
+++ b/src/compaction_filter.h
@@ -1,31 +1,32 @@
 #pragma once
 
+#include <string>
 #include <utility>
 
 #include "db_impl.h"
 #include "rocksdb/compaction_filter.h"
+#include "util/mutexlock.h"
 
 namespace rocksdb {
 namespace titandb {
 
 class TitanCompactionFilter final : public CompactionFilter {
  public:
-  explicit TitanCompactionFilter(
-      TitanDBImpl *db, const CompactionFilter *original,
-      std::unique_ptr<CompactionFilter> &&owned_filter,
-      std::shared_ptr<BlobStorage> blob_storage, bool skip_value)
+  TitanCompactionFilter(TitanDBImpl *db, const std::string &cf_name,
+                        const CompactionFilter *original,
+                        std::unique_ptr<CompactionFilter> &&owned_filter,
+                        std::shared_ptr<BlobStorage> blob_storage,
+                        bool skip_value)
       : db_(db),
+        cf_name_(cf_name),
         blob_storage_(std::move(blob_storage)),
         original_filter_(original),
         owned_filter_(std::move(owned_filter)),
-        skip_value_(skip_value) {
-    if (original_filter_ != nullptr) {
-      filter_name_ = std::string("TitanCompactionFilter.")
-                         .append(original_filter_->Name());
-    } else {
-      filter_name_ =
-          std::string("TitanCompactionFilter.").append(owned_filter_->Name());
-    }
+        skip_value_(skip_value),
+        filter_name_(std::string("TitanCompactionfilter.")
+                         .append(original_filter_->Name())) {
+    assert(blob_storage_ != nullptr);
+    assert(original_filter_ != nullptr);
   }
 
   const char *Name() const override { return filter_name_.c_str(); }
@@ -34,18 +35,12 @@ class TitanCompactionFilter final : public CompactionFilter {
                     const Slice &value, std::string *new_value,
                     std::string *skip_until) const override {
     if (skip_value_) {
-      return original_filter_ != nullptr
-                 ? original_filter_->FilterV2(level, key, value_type, Slice(),
-                                              new_value, skip_until)
-                 : owned_filter_->FilterV2(level, key, value_type, Slice(),
-                                           new_value, skip_until);
+      return original_filter_->FilterV2(level, key, value_type, Slice(),
+                                        new_value, skip_until);
     }
     if (value_type != kBlobIndex) {
-      return original_filter_ != nullptr
-                 ? original_filter_->FilterV2(level, key, value_type, value,
-                                              new_value, skip_until)
-                 : owned_filter_->FilterV2(level, key, value_type, value,
-                                           new_value, skip_until);
+      return original_filter_->FilterV2(level, key, value_type, value,
+                                        new_value, skip_until);
     }
 
     BlobIndex blob_index;
@@ -53,9 +48,13 @@ class TitanCompactionFilter final : public CompactionFilter {
     Status s = blob_index.DecodeFrom(&original_value);
     if (!s.ok()) {
       ROCKS_LOG_ERROR(db_->db_options_.info_log,
-                      "[%s] [key=%s] Unable to decode blob index", this->Name(),
-                      key.data());
-      db_->SetBGError(s);
+                      "[%s] Unable to decode blob index", cf_name_.c_str());
+      // TODO(yiwu): Better to fail the compaction as well, but current
+      // compaction filter API doesn't support it.
+      {
+        MutexLock l(&db_->mutex_);
+        db_->SetBGError(s);
+      }
       // Unable to decode blob index. Keeping the value.
       return Decision::kKeep;
     }
@@ -66,39 +65,36 @@ class TitanCompactionFilter final : public CompactionFilter {
 
     BlobRecord record;
     PinnableSlice buffer;
-
-    if (blob_storage_) {
-      ReadOptions read_options;
-      s = blob_storage_->Get(read_options, blob_index, &record, &buffer);
-    } else {
-      // kKeep is better. Maybe column family is not found due to some bugs.
-      return Decision::kKeep;
-    }
+    ReadOptions read_options;
+    s = blob_storage_->Get(read_options, blob_index, &record, &buffer);
 
     if (s.IsCorruption()) {
-      // meet a stale blob index, or bug. so just keep it
+      // Could be cause by blob file beinged GC-ed, or real corruption.
+      // TODO(yiwu): Tell the two cases apart.
       return Decision::kKeep;
     } else if (s.ok()) {
-      auto decision =
-          original_filter_ != nullptr
-              ? original_filter_->FilterV2(level, key, kValue, record.value,
-                                           new_value, skip_until)
-              : owned_filter_->FilterV2(level, key, kValue, record.value,
-                                        new_value, skip_until);
+      auto decision = original_filter_->FilterV2(
+          level, key, kValue, record.value, new_value, skip_until);
 
-      // It would be a problem if it change the value whereas the value_type is
-      // still kBlobIndex. For now, just returns kKeep.
+      // It would be a problem if it change the value whereas the value_type
+      // is still kBlobIndex. For now, just returns kKeep.
       // TODO: we should make rocksdb Filter API support changing value_type
       // assert(decision != CompactionFilter::Decision::kChangeValue);
       if (decision == Decision::kChangeValue) {
-        db_->SetBGError(Status::NotSupported(
-            "It would be a problem if it change the value whereas the "
-            "value_type is still kBlobIndex."));
+        {
+          MutexLock l(&db_->mutex_);
+          db_->SetBGError(Status::NotSupported(
+              "It would be a problem if it change the value whereas the "
+              "value_type is still kBlobIndex."));
+        }
         decision = Decision::kKeep;
       }
       return decision;
     } else {
-      db_->SetBGError(s);
+      {
+        MutexLock l(&db_->mutex_);
+        db_->SetBGError(s);
+      }
       // GetBlobRecord failed, keep the value.
       return Decision::kKeep;
     }
@@ -106,6 +102,7 @@ class TitanCompactionFilter final : public CompactionFilter {
 
  private:
   TitanDBImpl *db_;
+  const std::string cf_name_;
   std::shared_ptr<BlobStorage> blob_storage_;
   const CompactionFilter *original_filter_;
   const std::unique_ptr<CompactionFilter> owned_filter_;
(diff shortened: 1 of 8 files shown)


src/compaction_filter.h
```diff
@@ -0,0 +1,111 @@
...
+
+    BlobIndex blob_index;
+    Slice origin_value(value.data());
+    Status s = blob_index.DecodeFrom(&origin_value);
+    if (!s.ok()) {
+      // Unable to decode blob index. Keeping the value.
+      return Decision::kKeep;
+    }
+    if (BlobIndex::IsDeletionMarker(blob_index)) {
+      return Decision::kRemove;
```
yiwu-arbug, line 35: should keep the marker.

src/compaction_filter.h
```diff
@@ -0,0 +1,111 @@
...
+    if (value_type != kBlobIndex) {
+      return origin_filter_->FilterV2(level, key, value_type, value, new_value,
+                                      skip_until);
+    }
+
+    BlobIndex blob_index;
+    Slice origin_value(value.data());
+    Status s = blob_index.DecodeFrom(&origin_value);
+    if (!s.ok()) {
+      // Unable to decode blob index. Keeping the value.
```
Connor1996, line 31: maybe should set background error here? What do you think? @yiwu-arbug

src/compaction_filter.h
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    if (s.ok()) {
+      bool value_changed = false;
+      bool rv = origin_filter_->Filter(level, key, record.value, new_value,
+                                       &value_changed);
+      if (rv) {
+        return Decision::kRemove;
+      }
+      return value_changed ? Decision::kChangeValue : Decision::kKeep;
```
Connor1996, line 62: It would be a problem if you change the value whereas the value_type is still `kBlobIndex`. Seems we should make Filter API support changing value_type too? /cc @yiwu-arbug 

For now, we can just assert that `value_change == false` and abort. (Please also add a comment here).

src/compaction_filter.h
```diff
@@ -0,0 +1,111 @@
...
+      ReadOptions read_options;
+      s = storage->Get(read_options, blob_index, &record, &buffer);
+    } else {
+      // Column family not found, remove the value.
+      return Decision::kRemove;
+    }
+
+    if (s.ok()) {
+      bool value_changed = false;
+      bool rv = origin_filter_->Filter(level, key, record.value, new_value,
```
Connor1996, line 57: why not use `FilterV2`?

gotoxu: 这里调用`FilterV2`会产生和上述`value_changed == true`一样的问题，因此选择调用`Filter`。将来rocksdb本身的API可以支持`kBlobIndex`后可改为调用`FilterV2`

src/compaction_filter.h
gotoxu: 改为：
```cpp
assert(!value_changed);
return Decision::kKeep;
```

src/compaction_filter.h
gotoxu: What does **background error** mean? Write an error log? @Connor1996

src/compaction_filter.h
```diff
@@ -0,0 +1,115 @@
...
+      return Decision::kKeep;
+    }
+
+    // GetBlobRecord failed, keep the value.
+    return Decision::kKeep;
+  }
+
+private:
+  const TitanDBImpl *titan_db_impl_;
+  const CompactionFilter *origin_filter_;
```
yiwu-arbug, line 75: s/origin/original/

src/compaction_filter.h
```diff
@@ -0,0 +1,115 @@
...
+    if (BlobIndex::IsDeletionMarker(blob_index)) {
+      return Decision::kRemove;
+    }
+
+    BlobRecord record;
+    PinnableSlice buffer;
+
+    titan_db_impl_->mutex_.Lock();
+    auto storage =
+        titan_db_impl_->blob_file_set_->GetBlobStorage(column_family_id_)
```
yiwu-arbug, line 43: hmm, shall we cache `blob_storage` for each compaction filter instance? To avoid the need to lock global mutex for each key. The CF won't be dropped while compaction is still running.

src/compaction_filter.h
yiwu-arbug: rocksdb/titan will enter read only mode once it hits write error or compaction error. There's this flag in titan: https://github.com/tikv/titan/blob/master/src/db_impl.h#L282.

But since the compaction filter doesn't provide API to return its error, I think it is also fine to return `kKeep`.

src/compaction_filter.h
```diff
@@ -0,0 +1,115 @@
...
+
+    BlobIndex blob_index;
+    Slice origin_value(value.data());
+    Status s = blob_index.DecodeFrom(&origin_value);
+    if (!s.ok()) {
+      // Unable to decode blob index. Keeping the value.
+      return Decision::kKeep;
+    }
+    if (BlobIndex::IsDeletionMarker(blob_index)) {
+      return Decision::kRemove;
```
yiwu-arbug, line 35: Return `kKeep` here, and leave a TODO for me:
```
// TODO(yiwu): handle deletion marker at bottom level.
```

src/compaction_filter.h
yiwu-arbug: I still don't understand why `FilterV2` don't work. Can you explain a bit more?

src/compaction_filter.h
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    return std::unique_ptr<CompactionFilter>(new TitanCompactionFilter(
+        titan_db_impl_, compaction_filter, context.column_family_id));
+  }
+
+private:
+  const TitanDBImpl *titan_db_impl_;
+  const CompactionFilter *origin_filter_;
+  std::shared_ptr<CompactionFilterFactory> origin_cf_factory_;
```
yiwu-arbug, line 111: s/origin_cf_factory_/original_compaction_filter_factory_/

src/compaction_filter.h
```diff
@@ -0,0 +1,115 @@
...
+  uint32_t column_family_id_;
+};
+
+class TitanCompactionFilterFactory final : public CompactionFilterFactory {
+public:
+  explicit TitanCompactionFilterFactory(const TitanDBImpl *db)
+      : titan_db_impl_(db), origin_filter_(nullptr),
+        origin_cf_factory_(nullptr) {}
+
+  const char *Name() const override { return "TitanCompactionFilterFactory"; }
```
yiwu-arbug, line 85: can we return something like `"TitanCompactionFilterFactory."+original_filter_facotory.Name()`?

src/compaction_filter.h
```diff
@@ -0,0 +1,115 @@
...
+  void SetCFFactory(std::shared_ptr<CompactionFilterFactory> cf_factory) {
+    origin_cf_factory_ = std::move(cf_factory);
+  }
+
+  std::unique_ptr<CompactionFilter>
+  CreateCompactionFilter(const CompactionFilter::Context &context) override {
+    const CompactionFilter *compaction_filter = origin_filter_;
+    std::unique_ptr<CompactionFilter> compaction_filter_from_factory = nullptr;
+    if (compaction_filter == nullptr) {
+      // One of the origin_filter_ and origin_cf_factory_ must not be null
```
yiwu-arbug, line 98: shall we not make this assumption here?

src/compaction_filter.h
```diff
@@ -0,0 +1,115 @@
...
+
+class TitanCompactionFilterFactory final : public CompactionFilterFactory {
+public:
+  explicit TitanCompactionFilterFactory(const TitanDBImpl *db)
+      : titan_db_impl_(db), origin_filter_(nullptr),
+        origin_cf_factory_(nullptr) {}
+
+  const char *Name() const override { return "TitanCompactionFilterFactory"; }
+
+  void SetCF(const CompactionFilter *cf) { origin_filter_ = cf; }
```
yiwu-arbug, line 87: s/SetCF/SetOriginalCompactionFilter/

src/compaction_filter.h
```diff
@@ -0,0 +1,115 @@
...
+
+  void SetCFFactory(std::shared_ptr<CompactionFilterFactory> cf_factory) {
```
yiwu-arbug, line 89: s/SetCFFactory/SetOriginalCompactionFilterFactory/

src/db_impl.cc
```diff
@@ -269,6 +270,20 @@ Status TitanDBImpl::OpenImpl(const std::vector<TitanCFDescriptor>& descs,
         blob_file_set_.get(), stats_.get()));
     cf_opts.table_factory = titan_table_factories.back();
     cf_opts.merge_operator = shared_merge_operator_;
+    if (cf_opts.compaction_filter != nullptr ||
```
yiwu-arbug, line 273: We need to update the same logic to `CreateColumnFamilies()`

src/compaction_filter.h
gotoxu: Sorry，这里我理解有误，现在修改为：
```cpp
    if (s.ok()) {
      auto decision = origin_filter_->FilterV2(
          level, key, ValueType::kValue, record.value, new_value, skip_until);
      // It would be a problem if you change the value whereas the value_type is
      // still kBlobIndex. For now, we can just assert value_change == false and
      // abort.
      // TODO: we should make rocksdb Filter API support changing value_type
      assert(decision != CompactionFilter::Decision::kChangeValue);
      return decision;
    }
```

考虑到将来rocksdb的API可能会修改V2的逻辑，因此直接调用V2接口比较合适。

src/compaction_filter.h
gotoxu: Changed to original

src/compaction_filter.h
gotoxu:
```cpp
    if (BlobIndex::IsDeletionMarker(blob_index)) {
      // TODO(yiwu): handle deletion marker at bottom level.
      return Decision::kKeep;
    }
```

src/compaction_filter.h
gotoxu: Change to `original_filter_factory_`

src/compaction_filter.h
gotoxu: Change to: 
```cpp
if (original_filter_ != nullptr) {
      return std::string("TitanCompactionFilterFactory.")
          .append(original_filter_->Name())
          .c_str();
    } else if (original_filter_factory_ != nullptr) {
      return std::string("TitanCompactionFilterFactory.")
          .append(original_filter_factory_->Name())
          .c_str();
    } else {
      return "TitanCompactionFilterFactory.unknown";
    }
```

src/compaction_filter.h
gotoxu: if original_filter_ == null && original_filter_factory_ == null，那么TitanCompactionFilterFactory就是一个无意义的对象，现在在CreateCompactionFilter方法中加了一个推断：
```cpp
assert(original_filter_ != nullptr || original_filter_factory_ != nullptr);
```

src/db_impl.cc
gotoxu: Done

src/compaction_filter.h
gotoxu: Done!

src/compaction_filter.h
gotoxu: set `blob_storage` for each compaction filter instance in ctor.

src/compaction_filter.h
```diff
@@ -0,0 +1,129 @@
...
+      assert(decision != CompactionFilter::Decision::kChangeValue);
+      return decision;
+    }
+
+    // GetBlobRecord failed, keep the value.
+    return Decision::kKeep;
+  }
+
+private:
+  const std::weak_ptr<BlobStorage> blob_storage_;
```
yiwu-arbug, line 71: Just hold a `shared_ptr`. I'm wanting to cleanup the use of `weak_ptr` throughout Titan codebase.

src/compaction_filter.h
```diff
@@ -0,0 +1,129 @@
...
+  std::unique_ptr<CompactionFilter>
+  CreateCompactionFilter(const CompactionFilter::Context &context) override {
+    assert(original_filter_ != nullptr || original_filter_factory_ != nullptr);
+
+    const CompactionFilter *compaction_filter = original_filter_;
+    std::unique_ptr<CompactionFilter> compaction_filter_from_factory = nullptr;
+    if (compaction_filter == nullptr) {
+      compaction_filter_from_factory =
+          original_filter_factory_->CreateCompactionFilter(context);
+      compaction_filter = compaction_filter_from_factory.get();
```
yiwu-arbug, line 110: the original filter (the unique_ptr) is destroyed after the method call. The `TitanCompactionFilter` should hold the unique_ptr instead of raw pointer.

src/db_impl.h
```diff
@@ -167,6 +169,7 @@ class TitanDBImpl : public TitanDB {
   friend class BaseDbListener;
   friend class TitanDBTest;
   friend class TitanThreadSafetyTest;
+  friend class TitanCompactionFilterFactory;
```
yiwu-arbug, line 172: This is useless now?

src/compaction_filter.h
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    if (s.ok()) {
+      auto decision = original_filter_->FilterV2(
+          level, key, ValueType::kValue, record.value, new_value, skip_until);
+      // It would be a problem if you change the value whereas the value_type is
+      // still kBlobIndex. For now, we can just assert value_change == false and
+      // abort.
+      // TODO: we should make rocksdb Filter API support changing value_type
+      assert(decision != CompactionFilter::Decision::kChangeValue);
```
yiwu-arbug, line 62: hmmm, we should at least handle this one. Let's return `kKeep` if user compaction filter request to change the value.

Connor1996: but it maybe makes customized filter logic wrong and breaks data correctness.

yiwu-arbug: Agree. Let's set background error in this case then. Also set background error for blob index decode error above.

src/db_impl.h
gotoxu: `TitanCompactionFilterFactory` need to access `blob_file_set_` field in `CreateCompactionFilter` method.

src/compaction_filter.h
gotoxu:
```cpp
      if (decision == Decision::kChangeValue) {
        decision = Decision::kKeep;
      }
      return decision;
```

src/compaction_filter.h
gotoxu: Done!

gotoxu:
> How about we also add an option in TitanCFOptions to alternatively return empty value to compaction filter?

所以你的意思是，用户可以选择是否在Compaction Filter中传递value，以提高程序运行效率？

src/compaction_filter.h
```diff
@@ -53,62 +53,76 @@ class TitanCompactionFilter final : public CompactionFilter {
...
-      bool rv = origin_filter_->Filter(level, key, record.value, new_value,
-                                       &value_changed);
-      if (rv) {
-        return Decision::kRemove;
-      }
-      // It would be a problem if you change the value whereas the value_type is still kBlobIndex.
-      // For now, we can just assert value_change == false and abort.
+      auto decision = original_filter_->FilterV2(
+          level, key, ValueType::kValue, record.value, new_value, skip_until);
+      // It would be a problem if you change the value whereas the value_type is
```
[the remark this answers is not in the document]
gotoxu: Done!

src/compaction_filter.h
```diff
@@ -53,62 +53,76 @@ class TitanCompactionFilter final : public CompactionFilter {
...
+      // still kBlobIndex. For now, we can just assert value_change == false and
```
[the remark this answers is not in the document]
gotoxu: Done!

src/compaction_filter.h
```diff
@@ -53,62 +53,76 @@ class TitanCompactionFilter final : public CompactionFilter {
...
+      // abort.
```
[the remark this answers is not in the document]
gotoxu: Done!

src/compaction_filter.h
```diff
@@ -0,0 +1,142 @@
...
+      // TODO(yiwu): handle deletion marker at bottom level.
+      return Decision::kKeep;
+    }
+
+    BlobRecord record;
+    PinnableSlice buffer;
+
+    if (blob_storage_) {
+      ReadOptions read_options;
+      s = blob_storage_->Get(read_options, blob_index, &record, &buffer);
```
Connor1996, line 56: if the status is not okay(missing blob file). We can regard it as a stale index, and return `Decision::kRemove`

src/compaction_filter.h
```diff
@@ -0,0 +1,142 @@
...
+    } else {
+      // Column family not found, remove the value.
+      return Decision::kRemove;
```
Connor1996, line 59: `kKeep` is better. Maybe column family is not found due to some bugs.

src/compaction_filter.h
gotoxu: Did you mean `compaction_style`?

src/compaction_filter.h
```diff
@@ -0,0 +1,142 @@
...
+    if (value_type != kBlobIndex) {
+      return original_filter_->FilterV2(level, key, value_type, value,
+                                        new_value, skip_until);
+    }
+
+    BlobIndex blob_index;
+    Slice original_value(value.data());
+    Status s = blob_index.DecodeFrom(&original_value);
+    if (!s.ok()) {
+      // Unable to decode blob index. Keeping the value.
```
Connor1996, line 43: please call `db_->SetBGError(s)` here

src/compaction_filter.h
Connor1996: change logic like this
```suggestion
      s = blob_storage_->Get(read_options, blob_index, &record, &buffer);
      if (s.IsCorruption()) {
          // meet a stale blob index, or bug. so just keep it
          return Decision::kKeep;
      } else if (s.ok()) {
          ......
      } else {
          db_->SetBGError(s);
      }
```

src/compaction_filter.h
```diff
@@ -0,0 +1,142 @@
...
+
+    if (s.ok()) {
+      auto decision = original_filter_->FilterV2(
+          level, key, ValueType::kValue, record.value, new_value, skip_until);
+      // It would be a problem if it change the value whereas the value_type is
+      // still kBlobIndex. For now, just returns kKeep.
+      // TODO: we should make rocksdb Filter API support changing value_type
+      // assert(decision != CompactionFilter::Decision::kChangeValue);
+      if (decision == Decision::kChangeValue) {
+        decision = Decision::kKeep;
```
Connor1996, line 70: please call db_->SetBGError(s) here instead of setting to `Decision::kKeep`

src/compaction_filter_test.cc
```diff
@@ -0,0 +1,147 @@
...
+  ASSERT_OK(s);
+  ASSERT_EQ(value, "myvalue");
+
+  CompactAll();
+
+  s = Get("mykey", &value);
+  ASSERT_TRUE(s.IsNotFound());
+}
+
+TEST_F(TitanCompactionFilterTest, CompactBlobValue) {
```
Connor1996, line 121: Please add a case that you update a value.
```
Put(key, value1)  (value1 in blobfile1)
Put(key, value2)  (value2 in blobfile2)
... (for some reason, blobfile2 is GCed)
Compact();
Get(key)
```

src/compaction_filter.h
```diff
@@ -0,0 +1,152 @@
...
+      db_->SetBGError(s);
+      // GetBlobRecord failed, keep the value.
+      return Decision::kKeep;
+    }
+  }
+
+private:
+  TitanDBImpl *db_;
+  std::shared_ptr<BlobStorage> blob_storage_;
+  const CompactionFilter *original_filter_;
```
yiwu-arbug, line 91: how about this?
```
class TitanCompactionFilter {
 public:
  TitanCompactionFilter(..., CompactionFilter* original_filter, std::unique_ptr<CompactionFilter>&& owned_filter):
     original_filter_(original_filter), owned_filter_(std::move(owned_filter)) {}

  // keep existing usage of original_filter_.

 private:
  const CompactionFilter *original_filter_
  const std::unique_ptr<CompactionFilter> owned_filter_;
};
```

src/compaction_filter.h
yiwu-arbug: commented

src/compaction_filter.h
```diff
@@ -0,0 +1,152 @@
...
+    if (value_type != kBlobIndex) {
+      return original_filter_->FilterV2(level, key, value_type, value,
+                                        new_value, skip_until);
+    }
+
+    BlobIndex blob_index;
+    Slice original_value(value.data());
+    Status s = blob_index.DecodeFrom(&original_value);
+    if (!s.ok()) {
+      db_->SetBGError(s);
```
yiwu-arbug, line 44: print an info log using `ROCKS_LOG_ERROR` macro? you will need to pass `db_->db_options_->info_log` into `TitanCompactionFilter`.

src/compaction_filter.h
yiwu-arbug: I think it should be
```
if (s.ok()) {
   // return the value
} else if (s.IsNotFound()) {
   // ignore the value. The blob index has been GC-ed.
} else {
   db_->SetBGError(s);
}
```

yiwu-arbug: Unfortunately `Corruption` is an ambiguous error code where it could be a real corruption or blob file not found. We probably need to update `BlobStorage::Get` to return a flag to tell this two cases apart.

src/compaction_filter.h
```diff
@@ -0,0 +1,152 @@
...
+          level, key, ValueType::kValue, record.value, new_value, skip_until);
+      // It would be a problem if it change the value whereas the value_type is
+      // still kBlobIndex. For now, just returns kKeep.
+      // TODO: we should make rocksdb Filter API support changing value_type
+      // assert(decision != CompactionFilter::Decision::kChangeValue);
+      if (decision == Decision::kChangeValue) {
+        db_->SetBGError(Status::NotSupported(
+            "It would be a problem if it change the value whereas the "
+            "value_type is still kBlobIndex."));
+        // decision = Decision::kKeep;
```
yiwu-arbug, line 78: remove the commented-out line.

yiwu-arbug, line 78: I think we should return `kKeep`.

yiwu-arbug:
> > How about we also add an option in TitanCFOptions to alternatively return empty value to compaction filter?
> 
> 所以你的意思是，用户可以选择是否在Compaction Filter中传递value，以提高程序运行效率？

嗯，我其实觉得比较好的做法是修改用户逻辑，把 compaction filter 需要的 metadata 放在 key 里面。直接读 value 返回给 compaction filter 只是一个兼容模式。

src/compaction_filter_test.cc
```diff
@@ -0,0 +1,180 @@
...
+TEST_F(TitanCompactionFilterTest, CompactUpdateValue) {
+  options_.blob_file_discardable_ratio = 0.01;
+  options_.min_blob_size = 1;
+  options_.target_file_size_base = 1;
+  ASSERT_OK(Open());
+
+  ASSERT_OK(db_->Put(WriteOptions(), "update-key", "remain1"));
+  ASSERT_OK(db_->Put(WriteOptions(), "update-another-key", "remain2"));
+  ASSERT_OK(db_->Flush(FlushOptions()));
+  ASSERT_OK(db_->Put(WriteOptions(), "update-key", "remain3"));
```
Connor1996, line 158: let make it not to `remain3`, which result is expected for later get?

src/compaction_filter.h
gotoxu: So, 返回kKeep虽然有瑕疵，但是如果返回kChangeValue可能会导致一些问题，在目前还无法处理的情况下，返回kKeep并向用户说明不能修改value看起来是一个比较可行的方案!

gotoxu:
> > > How about we also add an option in TitanCFOptions to alternatively return empty value to compaction filter?
> > 
> > 
> > 所以你的意思是，用户可以选择是否在Compaction Filter中传递value，以提高程序运行效率？
> 
> 嗯，我其实觉得比较好的做法是修改用户逻辑，把 compaction filter 需要的 metadata 放在 key 里面。直接读 value 返回给 compaction filter 只是一个兼容模式。

How about this: 

options: `bool pass_value_in_compaction_filter`
default: `false`

```cpp
if (!pass_value_in_compaction_filter) {
    return user_compaction_filter->FilterV2(level, key, kValue, Slice(), new_value, skip_until);
} else {
    ...
}
```

src/compaction_filter_test.cc
gotoxu: I don't quite understand you mean. Can you be more specific?

yiwu-arbug:
> > > > How about we also add an option in TitanCFOptions to alternatively return empty value to compaction filter?
> > > 
> > > 
> > > 所以你的意思是，用户可以选择是否在Compaction Filter中传递value，以提高程序运行效率？
> > 
> > 
> > 嗯，我其实觉得比较好的做法是修改用户逻辑，把 compaction filter 需要的 metadata 放在 key 里面。直接读 value 返回给 compaction filter 只是一个兼容模式。
> 
> How about this:
> 
> options: `bool pass_value_in_compaction_filter`
> default: `false`
> 
> ```c++
> if (!pass_value_in_compaction_filter) {
>     return user_compaction_filter->FilterV2(level, key, kValue, Slice(), new_value, skip_until);
> } else {
>     ...
> }
> ```

looking good, except s/pass/skip/

src/compaction_filter_test.cc
Connor1996:
```suggestion
  ASSERT_OK(db_->Put(WriteOptions(), "update-key", "value"));
```
like this, just to make it not removed by compaction filter

yiwu-arbug:
> Oh, it's still a problem that compaction filter deletes a blob index in the upper level which exposes the stale blob index.
> 
> I have no better idea of this except assuming the user won't read the key again. Seems it's a common issue for rocksdb compaction filter? /cc @yiwu-arbug

When compaction filter remove a value, the value is replace by a delete tombstone, so stale values won't reappear: https://github.com/facebook/rocksdb/blob/master/db/compaction/compaction_iterator.cc#L219

src/compaction_filter_test.cc
gotoxu: But now `remain3` will not be removed by compaction filter! Change to `value` will be removed by CF.

Connor1996: sorry, it's a typo. I mean making it removed by compaction filter, so we can ensure the stale version is not exposed.

src/compaction_filter.h
```diff
@@ -0,0 +1,174 @@
...
+      return std::string("TitanCompactionFilter.")
+          .append(owned_filter_->Name())
+          .c_str();
+    }
+  }
+
+  Decision FilterV2(int level, const Slice &key, ValueType value_type,
+                    const Slice &value, std::string *new_value,
+                    std::string *skip_until) const override {
+    if (skip_value_) {
```
Connor1996, line 36: move it after `if (value_type != kBlobIndex) { ... }` and pass the `value_type` instead of `kValue`.

gotoxu: 1. 如果把这段代码移到`if (value_type != kBlobIndex)`判断之后，那么该配置表示的含义则变成只有当`value_type ==kBlobIndex`时才会生效，这样是否合理？@yiwu-arbug

2. 即使将这段代码移到`if (value_type != kBlobIndex)`判断之后，那么传递的`value_type`也只能是`kValue`，因为此时的`value_type == kBlobIndex`，如果直接传入`FilterV2`中，那么执行逻辑就是直接返回`Decision::kKeep`，这应该不是我们想要的结果。

yiwu-arbug: I prefer returning empty regardless of whether it is value or blob index. It is more consistent and make it easier for compaction filter developer to notice the behavior. And 2 is a good point, too. We probably still want to pass `kMergeOperand` along though, not replacing it with `kValue`.

gotoxu:
Why the code that I formatted using the local clang-format-diff script fails the check in CI?

yiwu-arbug:
> Why the code that I formatted using the local clang-format-diff script fails the check in CI?

not sure. CI use the same script.

gotoxu:
I ran `scripts/format-diff.sh` @yiwu-arbug

src/compaction_filter.h
```diff
@@ -0,0 +1,171 @@
...
+ }
+
+ const char *Name() const override { return filter_name_.c_str(); }
+
+ Decision FilterV2(int level, const Slice &key, ValueType value_type,
+                   const Slice &value, std::string *new_value,
+                   std::string *skip_until) const override {
+   if (skip_value_) {
+     return original_filter_ != nullptr
+                ? original_filter_->FilterV2(level, key, kValue, Slice(),
```
Connor1996, line 38:
```suggestion
                ? original_filter_->FilterV2(level, key, value_type, Slice(),
```

src/compaction_filter.h
```diff
@@ -0,0 +1,171 @@
...
+                                             new_value, skip_until)
+                : owned_filter_->FilterV2(level, key, kValue, Slice(),
```
Connor1996, line 40:
```suggestion
                : owned_filter_->FilterV2(level, key, value_type, Slice(),
```

gotoxu: Sorry, forget to change this.

yiwu-arbug:
clang-format appear to work on my side. I also updated the PR with some smaller changes. Will defer to @Connor1996 for merge.

gotoxu:
Thanks. 一旦这个PR合并到仓库中，大概多久能应用到TiKV中？是否有什么更新的计划么？

yiwu-arbug:
> Thanks. 一旦这个PR合并到仓库中，大概多久能应用到TiKV中？是否有什么更新的计划么？

你在用哪个版本的tikv，3.0.x吗？

gotoxu:
> > Thanks. 一旦这个PR合并到仓库中，大概多久能应用到TiKV中？是否有什么更新的计划么？
> 
> 你在用哪个版本的tikv，3.0.x吗？

release-4.0 这个分支

yiwu-arbug:
> > > Thanks. 一旦这个PR合并到仓库中，大概多久能应用到TiKV中？是否有什么更新的计划么？
> > 
> > 
> > 你在用哪个版本的tikv，3.0.x吗？
> 
> release-4.0 这个分支

争取周五之前弄进去这个分支，4.0下周有一个小版本。

gotoxu:
> > > > Thanks. 一旦这个PR合并到仓库中，大概多久能应用到TiKV中？是否有什么更新的计划么？
> > > 
> > > 
> > > 你在用哪个版本的tikv，3.0.x吗？
> > 
> > 
> > release-4.0 这个分支
> 
> 争取周五之前弄进去这个分支，4.0下周有一个小版本。

Thanks!!!

Connor1996:
tests failed, please fix it.

yiwu-arbug:
@gotoxu this change has been merged into tikv release-4.0 branch by https://github.com/tikv/tikv/pull/8038.

yiwu-arbug:
@gotoxu by the way, you can join tikv slack (tikv-wg.slack.com) and join the #sig-engine channel. if you have other Titan question and feature request, you can ping us there.

PR #168, opened 2020-05-29
Fix GC may delete a already deleted blob file

Connor1996:
Fix #167

PR #169, opened 2020-05-29
Fix GC may delete a already deleted blob file (#168)

Connor1996:
Cherry-pick of #168


Connor1996:
test should pass after #172 is merged

PR #172, opened 2020-06-02
Fix wrong assert delta < 0 for cocurrent compaction while flush

### LINKED ISSUE #171: titan_thread_safety_test failed
```
10/14 Test #10: titan_thread_safety_test ...............Child aborted***Exception:   5.00 sec
[==========] Running 2 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 2 tests from TitanThreadSafetyTest
[ RUN      ] TitanThreadSafetyTest.Insert
[       OK ] TitanThreadSafetyTest.Insert (4173 ms)
[ RUN      ] TitanThreadSafetyTest.InsertAndDelete
Assertion failed: (false), function OnCompactionCompleted, file /Users/travis/build/tikv/titan/src/db_impl.cc, line 1239.
```

occurs in #169

Connor1996:
Because the flushed SST is added to superversion before `OnFlushCompleted()` is called, so if there is a concurrent compaction, `OnCompactionCompleted()` maybe called before `OnFlushCompleted()` is called.
In this case, the state of the blob file generated by the flush is still `kPendingLSM`, while the blob file size delta is for the compaction event. So it is possible that delta is negative.

Fix #171


src/db_impl.cc
```diff
@@ -1230,14 +1243,29 @@ void TitanDBImpl::OnCompactionCompleted(
...
+          // positive number. And the delta will be subtracted in
+          // `OnFlushCompleted()` later.
+          SubStats(stats_.get(), compaction_job_info.cf_id,
+                   file->GetDiscardableRatioLevel(), 1);
+          file->set_live_data_size(static_cast<uint64_t>(-delta));
+          AddStats(stats_.get(), compaction_job_info.cf_id,
+                   file->GetDiscardableRatioLevel(), 1);
+          ROCKS_LOG_INFO(db_options_.info_log,
+                         "OnCompactionCompleted[%d]: Get blob file %" PRIu64
+                         " live size being negative %lld, maybe due to "
```
yiwu-arbug, line 1265: use PRIu64 macro instead of `%lld`

src/db_impl.cc
```diff
@@ -1148,6 +1149,18 @@ void TitanDBImpl::OnFlushCompleted(const FlushJobInfo& flush_job_info) {
...
+        // Because the flushed SST is added to superversion before
+        // `OnFlushCompleted()` is called, so if there is a concurrent
+        // compaction, `OnCompactionCompleted()` maybe called before
+        // `OnFlushCompleted()` is called.
+        // In this case, the state of the blob file generated by the flush is
+        // still `kPendingLSM`, while the blob file size delta is for the
+        // compaction event. So it is possible that delta is negative.
+        // It records the delta as a positive number if any, so here subtract it
+        // from the total live data size.
+        delta -= file->live_data_size();
```
yiwu-arbug, line 1162: Also add some comment in blob_format.h around `BlobFileMeta::live_data_size` to explain what does the value mean when `state_ == kPendingLSM`?

src/db_impl.cc
```diff
@@ -1230,14 +1243,29 @@ void TitanDBImpl::OnCompactionCompleted(
...
-        if (delta < 0) {
-          // Cannot happen..
-          ROCKS_LOG_WARN(db_options_.info_log,
-                         "OnCompactionCompleted[%d]: New blob file %" PRIu64
-                         " live size being negative",
-                         compaction_job_info.job_id, file_number);
-          assert(false);
-          delta = 0;
+        if (delta <= 0) {
+          // Because the flushed SST is added to superversion before
```
yiwu-arbug, line 1247: It seems to me the same could happen to compaction-generated blob files. Update the comment to reflect that?

src/db_impl.cc
```diff
@@ -1230,14 +1243,29 @@ void TitanDBImpl::OnCompactionCompleted(
...
+          // so if there is a concurrent compaction, `OnCompactionCompleted()`
+          // maybe called before `OnFlushCompleted()` is called.
+          // In this case, the state of the blob file generated by the flush is
+          // still `kPendingLSM`, while the blob file size delta
+          // is for the compaction event. So it is possible that delta is
+          // negative.
+          // To make the live data size accurate, here records the delta as a
+          // positive number. And the delta will be subtracted in
+          // `OnFlushCompleted()` later.
+          SubStats(stats_.get(), compaction_job_info.cf_id,
```
yiwu-arbug, line 1258: Maybe we shouldn't add discardable-ratio-level stat when the file is still in `kPendingLSM` state?

Connor1996: the discardable-ratio-level is updated when the blob file is added in manifest. But at that time the live_data_size is unknown, so here we need to update the discardable-ratio-level stat.

Connor1996: we have talked about it in this PR https://github.com/tikv/titan/pull/157

yiwu-arbug: lol, sorry, I feel it weird every time.

Connor1996: indeed, will add a todo here.

Connor1996: make it not count zero live data size when adding blob file

PR #189, opened 2020-09-26
Add dictionary compression support for blob file

### LINKED ISSUE #118: PCP-23-2: Build blob file using zstd dictionary compression
This is subtask of https://github.com/tikv/tikv/issues/5743

Add `CompressionOptions blob_file_compression_options` to `TitanCFOptions`. `BlobFileBuilder` should respect `blob_file_compression_options.enabled`. Also if `max_dict_bytes > 0`, encode the content using zstd dictionary. If `zstd_max_train_bytes = 0`, it should use all the content as samples to train dictionary; otherwise, the first `zstd_max_train_bytes` is used. The samples should be store in memory before the dictionary is trained, and then flush to file after that. The dictionary will be stored in blob file as meta block.

For now, update `BlobFileReader` and `BlobFileIterator` to return `Status::NotSupported` if they encounter a file come with dictionary compression. We will update them in the next task.

Update table_builder_test to add tests for the change.

chux0519:
Fixes #118

Briefly I've done following things:
- add compression options in TitanCFOptions
- set flags in header when there is compression dictionary
- add some members(states, sample data, etc.) in BlobFileBuilder for dictionary compression logic
- write the dictionary in meta blocks, when dictionary compression is enabled
- refactor BlobEncoder to support dictionary update

currently for decoding
- update BlobFileReader and BlobFileIterator to return Status::NotSupported if a blob file come with dictionary compression.
- add a test case for BlobFileReader

Issue Number: https://github.com/tikv/tikv/issues/8635

### DIFF
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -85,6 +85,11 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // Default: kNoCompression
   CompressionType blob_file_compression{kNoCompression};
 
+  // The compression options
+  //
+  // Default: disabled
+  CompressionOptions blob_file_compression_options;
+
   // The desirable blob file size. This is not a hard limit but a wish.
   //
   // Default: 256MB
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -85,9 +85,10 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // Default: kNoCompression
   CompressionType blob_file_compression{kNoCompression};
 
-  // The compression options
-  //
-  // Default: disabled
+  // The compression options. The `blob_file_compression.enabled` option is
+  // ignored, we only use `blob_file_compression` above to determine wether the
+  // blob file is compressed. We use this options mainly to configure the
+  // compression dictionary.
   CompressionOptions blob_file_compression_options;
 
   // The desirable blob file size. This is not a hard limit but a wish.
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -11,7 +11,8 @@ BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
                          : BuilderState::kUnbuffered),
       cf_options_(cf_options),
       file_(file),
-      encoder_(cf_options_.blob_file_compression) {
+      encoder_(cf_options_.blob_file_compression,
+               cf_options.blob_file_compression_options) {
   BlobFileHeader header;
   if (cf_options.blob_file_compression_options.max_dict_bytes > 0)
     header.flags |= BlobFileHeader::kHasUncompressionDictionary;
@@ -23,7 +24,9 @@ BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
 void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
   if (!ok()) return;
   if (builder_state_ == BuilderState::kBuffered) {
-    sample_records_.emplace_back(record);
+    std::string rec_str;
+    record.EncodeTo(&rec_str);
+    sample_records_.emplace_back(rec_str);
     sample_str_len_ += (16 /* 2 extra Varint64 */ + record.size());
     if (cf_options_.blob_file_compression_options.zstd_max_train_bytes > 0 &&
         sample_str_len_ >=
@@ -62,36 +65,36 @@ void BlobFileBuilder::EnterUnbuffered() {
   // When above things are done, transform builder state into unbuffered
   std::string samples = "";
   std::vector<size_t> sample_lens;
-  std::string rec_str;
 
-  const size_t kSampleBytes =
-      cf_options_.blob_file_compression_options.zstd_max_train_bytes > 0
-          ? cf_options_.blob_file_compression_options.zstd_max_train_bytes
-          : cf_options_.blob_file_compression_options.max_dict_bytes;
-  for (const auto& rec : sample_records_) {
-    rec.EncodeTo(&rec_str);
-    size_t copy_len = std::min(kSampleBytes - samples.size(), rec_str.size());
+  for (const auto& rec_str : sample_records_) {
+    size_t copy_len =
+        cf_options_.blob_file_compression_options.zstd_max_train_bytes > 0
+            ? std::min(cf_options_.blob_file_compression_options
+                               .zstd_max_train_bytes -
+                           rec_str.size(),
+                       rec_str.size())
+            : rec_str.size();
     samples.append(rec_str, 0, copy_len);
     sample_lens.emplace_back(copy_len);
   }
   std::string dict;
-  if (cf_options_.blob_file_compression_options.zstd_max_train_bytes > 0) {
-    dict = ZSTD_TrainDictionary(
-        samples, sample_lens,
-        cf_options_.blob_file_compression_options.max_dict_bytes);
-  } else {
-    dict = std::move(samples);
-  }
-  CompressionDict compression_dict(
-      dict, cf_options_.blob_file_compression,
-      cf_options_.blob_file_compression_options.level);
-  encoder_.SetCompressionDict(compression_dict);
+  dict = ZSTD_TrainDictionary(
+      samples, sample_lens,
+      cf_options_.blob_file_compression_options.max_dict_bytes);
+
+  compression_dict_.reset(
+      new CompressionDict(dict, cf_options_.blob_file_compression,
+                          cf_options_.blob_file_compression_options.level));
+  encoder_.SetCompressionDict(*compression_dict_);
 
   builder_state_ = BuilderState::kUnbuffered;
 
   // add history buffer
   BlobHandle handle;
-  for (const BlobRecord& rec : sample_records_) {
+  for (const std::string& rec_str : sample_records_) {
+    BlobRecord rec;
+    Slice rec_slice(rec_str);
+    rec.DecodeFrom(&rec_slice);
     Add(rec, &handle);
   }
   sample_records_.clear();
@@ -106,7 +109,7 @@ void BlobFileBuilder::WriteRawBlock(const Slice& block, BlockHandle* handle) {
 
 void BlobFileBuilder::WriteCompressionDictBlock(
     MetaIndexBuilder* meta_index_builder, BlockHandle* handle) {
-  WriteRawBlock(encoder_.GetCompressionDict(), handle);
+  WriteRawBlock(compression_dict_->GetRawDict(), handle);
   if (ok()) {
     meta_index_builder->Add(kCompressionDictBlock, *handle);
   }
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -19,19 +19,25 @@ BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
 }
 
 void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
-  if (!ok())
-    return;
+  if (!ok()) return;
   if (builder_state_ == BuilderState::kBuffered) {
     sample_records_.emplace_back(record);
     sample_str_len_ += (16 /* 2 extra Varint64 */ + record.size());
     if (cf_options_.blob_file_compression_options.zstd_max_train_bytes > 0 &&
         sample_str_len_ >=
             cf_options_.blob_file_compression_options.zstd_max_train_bytes) {
-      // TODO: should enter unbuffered state
+      EnterUnbuffered();
+      // add history buffer
+      for (const BlobRecord& rec : sample_records_) {
+        Add(rec, handle);
+      }
+      sample_records_.clear();
+      sample_str_len_ = 0;
     }
     return;
   }
 
+  assert(builder_state_ == BuilderState::kUnbuffered);
   // unbuffered state
   encoder_.EncodeRecord(record);
   handle->offset = file_->GetFileSize();
@@ -54,9 +60,42 @@ void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
   }
 }
 
+void BlobFileBuilder::EnterUnbuffered() {
+  // Using collected samples to train the compression dictionary
+  // Then replay those records in memory, encode them to blob file
+  // When above things are done, transform builder state into unbuffered
+  std::string samples = "";
+  std::vector<size_t> sample_lens;
+  std::string rec_str;
+
+  const size_t kSampleBytes =
+      cf_options_.blob_file_compression_options.zstd_max_train_bytes > 0
+          ? cf_options_.blob_file_compression_options.zstd_max_train_bytes
+          : cf_options_.blob_file_compression_options.max_dict_bytes;
+  for (const auto& rec : sample_records_) {
+    rec.EncodeTo(&rec_str);
+    size_t copy_len = std::min(kSampleBytes - samples.size(), rec_str.size());
+    samples.append(rec_str, 0, copy_len);
+    sample_lens.emplace_back(copy_len);
+  }
+  std::string dict;
+  if (cf_options_.blob_file_compression_options.zstd_max_train_bytes > 0) {
+    dict = ZSTD_TrainDictionary(
+        samples, sample_lens,
+        cf_options_.blob_file_compression_options.max_dict_bytes);
+  } else {
+    dict = std::move(samples);
+  }
+  CompressionDict compression_dict(
+      dict, cf_options_.blob_file_compression,
+      cf_options_.blob_file_compression_options.level);
+  encoder_.SetCompressionDict(compression_dict);
+
+  builder_state_ = BuilderState::kUnbuffered;
+}
+
 Status BlobFileBuilder::Finish() {
-  if (!ok())
-    return status();
+  if (!ok()) return status();
 
   std::string buffer;
   BlobFileFooter footer;
@@ -72,9 +111,7 @@ Status BlobFileBuilder::Finish() {
 
 void BlobFileBuilder::Abandon() {}
 
-uint64_t BlobFileBuilder::NumEntries() {
-  return num_entries_;
-}
+uint64_t BlobFileBuilder::NumEntries() { return num_entries_; }
 
 }  // namespace titandb
 }  // namespace rocksdb
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -91,7 +91,7 @@ void BlobFileBuilder::EnterUnbuffered() {
 void BlobFileBuilder::FlushSampleRecords() {
   BlobHandle handle;
   for (const std::string& record_str : sample_records_) {
-    encoder_.EncodeString(record_str);
+    encoder_.EncodeSlice(record_str);
     WriteEncoderData(&handle);
   }
   sample_records_.clear();
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -27,6 +27,9 @@ void BlobFileBuilder::Add(const BlobRecord& record,
   if (!ok()) return;
   std::string key = record.key.ToString();
   if (builder_state_ == BuilderState::kBuffered) {
+    // Check flags, if it is a small KV pair, do nothing, just cache it
+    if (ctx->small_kv_ctx.is_small)
+      return cached_contexts_.emplace_back(std::move(ctx));
     std::string record_str;
     // Encode to take ownership of underlying string.
     record.EncodeTo(&record_str);
@@ -39,6 +42,10 @@ void BlobFileBuilder::Add(const BlobRecord& record,
       EnterUnbuffered(out_ctx);
     }
   } else if (builder_state_ == BuilderState::kUnbuffered) {
+    // Check flags, if it is a small KV pair, do nothing, just return it
+    // back, and it will be written into sst out side
+    if (ctx->small_kv_ctx.is_small)
+      return out_ctx->emplace_back(std::move(ctx));
     encoder_.EncodeRecord(record);
     WriteEncoderData(&ctx->new_blob_index.blob_handle);
     out_ctx->emplace_back(std::move(ctx));
@@ -96,14 +103,23 @@ void BlobFileBuilder::EnterUnbuffered(OutContexts* out_ctx) {
 }
 
 void BlobFileBuilder::FlushSampleRecords(OutContexts* out_ctx) {
-  assert(cached_contexts_.size() == sample_records_.size());
-  for (size_t i = 0; i < sample_records_.size(); i++) {
+  assert(cached_contexts_.size() >= sample_records_.size());
+  size_t i, j;
+  for (i = 0; i < sample_records_.size(); i++, j++) {
     const std::string& record_str = sample_records_[i];
-    const std::unique_ptr<BlobRecordContext>& ctx = cached_contexts_[i];
+    j = i;
+    for (; j < cached_contexts_.size() &&
+           cached_contexts_[j]->small_kv_ctx.is_small;
+         j++) {
+      out_ctx->emplace_back(std::move(cached_contexts_[j]));
+    }
+    const std::unique_ptr<BlobRecordContext>& ctx = cached_contexts_[j];
     encoder_.EncodeSlice(record_str);
     WriteEncoderData(&ctx->new_blob_index.blob_handle);
-    out_ctx->emplace_back(std::move(cached_contexts_[i]));
+    out_ctx->emplace_back(std::move(cached_contexts_[j]));
   }
+  assert(i == sample_records_.size());
+  assert(j == cached_contexts_.size());
   sample_records_.clear();
   sample_str_len_ = 0;
   cached_contexts_.clear();
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -114,6 +114,20 @@ void BlobFileBuilder::WriteRawBlock(const Slice& block, BlockHandle* handle) {
   handle->set_offset(file_->GetFileSize());
   handle->set_size(block.size());
   status_ = file_->Append(block);
+  if (ok()) {
+    // follow rocksdb's block based table format
+    char trailer[kBlockTrailerSize];
+    // only compression dictionary and meta index block are written
+    // by this method, we use `kNoCompression` as placeholder
+    trailer[0] = kNoCompression;
+    char* trailer_without_type = trailer + 1;
+
+    // crc32 checksum
+    auto crc = crc32c::Value(block.data(), block.size());
+    crc = crc32c::Extend(crc, trailer, 1);  // Extend to cover block type
+    EncodeFixed32(trailer_without_type, crc32c::Mask(crc));
+    status_ = file_->Append(Slice(trailer, kBlockTrailerSize));
+  }
 }
 
 void BlobFileBuilder::WriteCompressionDictBlock(
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -64,13 +64,19 @@ void BlobFileBuilder::EnterUnbuffered(OutContexts* out_ctx) {
   std::vector<size_t> sample_lens;
 
   for (const auto& record_str : sample_records_) {
+    if (cf_options_.blob_file_compression_options.zstd_max_train_bytes > 0 &&
+        samples.size() >=
+            cf_options_.blob_file_compression_options.zstd_max_train_bytes) {
+      // Enough training data
+      break;
+    }
     size_t copy_len =
-        cf_options_.blob_file_compression_options.zstd_max_train_bytes > 0
-            ? std::min(cf_options_.blob_file_compression_options
+        cf_options_.blob_file_compression_options.zstd_max_train_bytes == 0
+            ? record_str.size()
+            : std::min(cf_options_.blob_file_compression_options
                                .zstd_max_train_bytes -
-                           record_str.size(),
-                       record_str.size())
-            : record_str.size();
+                           samples.size(),
+                       record_str.size());
     samples.append(record_str, 0, copy_len);
     sample_lens.emplace_back(copy_len);
   }
@@ -84,9 +90,9 @@ void BlobFileBuilder::EnterUnbuffered(OutContexts* out_ctx) {
                           cf_options_.blob_file_compression_options.level));
   encoder_.SetCompressionDict(*compression_dict_);
 
-  builder_state_ = BuilderState::kUnbuffered;
+  FlushSampleRecords(out_ctx);
 
-  return FlushSampleRecords(out_ctx);
+  builder_state_ = BuilderState::kUnbuffered;
 }
 
 void BlobFileBuilder::FlushSampleRecords(OutContexts* out_ctx) {
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -124,7 +124,7 @@ void BlobFileBuilder::WriteRawBlock(const Slice& block, BlockHandle* handle) {
 
     // crc32 checksum
     auto crc = crc32c::Value(block.data(), block.size());
-    crc = crc32c::Extend(crc, trailer, 1);  // Extend to cover block type
+    crc = crc32c::Extend(crc, trailer, 1);  // Extend to cover compression type
     EncodeFixed32(trailer_without_type, crc32c::Mask(crc));
     status_ = file_->Append(Slice(trailer, kBlockTrailerSize));
   }
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -147,14 +147,18 @@ void BlobFileBuilder::WriteCompressionDictBlock(
   }
 }
 
-// TODO: return indices as well
-Status BlobFileBuilder::Finish() {
-  if (!ok()) return status();
+BlobIndices BlobFileBuilder::Finish(Status* finish_status) {
+  BlobIndices ret;
+  if (!ok()) {
+    *finish_status = status();
+    return ret;
+  }
 
-  if (builder_state_ == BuilderState::kBuffered) EnterUnbuffered();
+  if (builder_state_ == BuilderState::kBuffered) {
+    ret = EnterUnbuffered();
+  }
 
   BlobFileFooter footer;
-
   // if has compression dictionary, encode it into meta blocks
   // and update relative fields in footer
   if (cf_options_.blob_file_compression_options.max_dict_bytes > 0) {
@@ -174,7 +178,8 @@ Status BlobFileBuilder::Finish() {
     // The Sync will be done in `BatchFinishFiles`
     status_ = file_->Flush();
   }
-  return status();
+  *finish_status = status();
+  return ret;
 }
 
 void BlobFileBuilder::Abandon() {}
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -13,13 +13,25 @@ BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
       file_(file),
       encoder_(cf_options_.blob_file_compression,
                cf_options.blob_file_compression_options) {
+  if (file_) WriteHeader();
+}
+
+void BlobFileBuilder::WriteHeader() {
   BlobFileHeader header;
-  if (cf_options.blob_file_compression_options.max_dict_bytes > 0)
+  if (cf_options_.blob_file_compression_options.max_dict_bytes > 0)
     header.flags |= BlobFileHeader::kHasUncompressionDictionary;
   std::string buffer;
   header.EncodeTo(&buffer);
   status_ = file_->Append(buffer);
 }
+
+void BlobFileBuilder::SetFileWriter(WritableFileWriter* writer) {
+  assert(file_ == nullptr);
+  file_ = writer;
+  if (file_) WriteHeader();
+}
+
+bool BlobFileBuilder::HasFileWriter() { return file_ != nullptr; }
 
 void BlobFileBuilder::Add(const BlobRecord& record,
                           std::unique_ptr<BlobRecordContext> ctx,
@@ -166,9 +178,9 @@ void BlobFileBuilder::WriteCompressionDictBlock(
 }
 
 Status BlobFileBuilder::Finish(OutContexts* out_ctx) {
-  if (!ok()) {
-    return status();
-  }
+  if (!HasFileWriter()) return status();
+
+  if (!ok()) return status();
 
   if (builder_state_ == BuilderState::kBuffered) {
     EnterUnbuffered(out_ctx);
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -37,27 +37,21 @@ void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
     return;
   }
 
-  assert(builder_state_ == BuilderState::kUnbuffered);
-  // unbuffered state
-  encoder_.EncodeRecord(record);
-  handle->offset = file_->GetFileSize();
-  handle->size = encoder_.GetEncodedSize();
-  live_data_size_ += handle->size;
+  if (builder_state_ == BuilderState::kUnbuffered) {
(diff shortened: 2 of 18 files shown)


Connor1996 (commented):
Please give it a meaningful title, and describe what you do in the PR description.

Connor1996 (commented):
please fix DCO

src/blob_file_builder.cc
```diff
@@ -6,18 +6,38 @@ namespace titandb {
...
+  if (cf_options.blob_file_compression_options.max_dict_bytes > 0)
+    header.flags |= BlobFileHeader::kHasUncompressionDictionary;
   std::string buffer;
   header.EncodeTo(&buffer);
   status_ = file_->Append(buffer);
 }
 
 void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
   if (!ok()) return;
+  if (builder_state_ == BuilderState::kBuffered) {
```
Connor1996, line 26: should update `smallest_key_` and `largest_key_`

Connor1996: forget it. Seems these will be updated correctly in `EnterUnbuffered`

src/blob_file_builder.cc
```diff
@@ -6,18 +6,38 @@ namespace titandb {
...
+    std::string rec_str;
+    record.EncodeTo(&rec_str);
+    sample_records_.emplace_back(rec_str);
+    sample_str_len_ += rec_str.size();
+    if (cf_options_.blob_file_compression_options.zstd_max_train_bytes > 0 &&
+        sample_str_len_ >=
+            cf_options_.blob_file_compression_options.zstd_max_train_bytes) {
+      EnterUnbuffered();
+    }
+    return;
```
Connor1996, line 36: `BlobGCJob::DoRun` will call `BlobFileBuilder::Add` and use the `BlobHandle` passed in. You should handle it correctly when in buffered state.

src/blob_file_builder.cc
```diff
@@ -6,18 +6,38 @@ namespace titandb {
...
   header.EncodeTo(&buffer);
   status_ = file_->Append(buffer);
 }
 
 void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
   if (!ok()) return;
+  if (builder_state_ == BuilderState::kBuffered) {
+    std::string rec_str;
+    record.EncodeTo(&rec_str);
+    sample_records_.emplace_back(rec_str);
```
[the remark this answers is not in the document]
chux0519: Thanks for reviewing! Actually I tried this at the first attempt, but the record value will be incorrect when popped out.
Say, if I use following code, when popped from sample_records_, the key of all records will become `key-9` which is incorrect. I guess the key slice is shared from somewhere, and is overwritten by later keys(maybe), so it supposed be copied here.

I choose to use a string to store it here, and decode it when entering unbuffered state.

```c++
for (int i = 0; i < 10; i++) {
    s = db->Put(opts, "key-" + std::to_string(i), "value-" + std::to_string(i));
 }
```

src/blob_file_builder.cc
chux0519: OK, I will take a look at this part, I'm still new to titan, and definitely missed this detail. thanks for your reminder, will fix this later!

chux0519: I took a look roughly. The value of `handle` was not set when `Add` was called in the `buffered` state.
And, the offset and size of the `handle` can not be calculated in advance.
So currently, my idea is to use a callback-style design.
1. `BlobFileBuilder` holds a callback queue, when calling `Add` in buffered state, add an callable object to the callback queue, then call it(with related handle value) when the corresponding record is actually compressed and written.
2. When calling `Add` from `BlobGCJob` and `TitanTableBuilder`, check the state of the builder, if it is in buffered state, using callback to do the rest of work.


Is this design reasonable? Do you have any suggestions or points that I should be attention? Thanks.

include/titan/options.h
```diff
@@ -85,6 +85,11 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // Default: kNoCompression
   CompressionType blob_file_compression{kNoCompression};
 
+  // The compression options
+  //
+  // Default: disabled
```
yiwu-arbug, line 90: remove this line.

src/blob_file_builder.cc
Connor1996: I think we can buffer the key first, and returned the blob handle by `BlobFileBuidler->Finish` and write the key and handle at that time

Connor1996: Also, please note `blob_builder_->Add` in `TitanTableBuilder::AddBlob` suffers from the same issue. Hmmmm, seems the code changes would be far more than I expected.

src/blob_file_builder.cc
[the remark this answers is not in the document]
Connor1996: Let's deep copy the slice and then push_back to sample_records

include/titan/options.h
```diff
@@ -85,6 +85,11 @@ struct TitanCFOptions : public ColumnFamilyOptions {
...
+  CompressionOptions blob_file_compression_options;
```
yiwu-arbug, line 91: Default disable would be confusing, when `blob_file_compression` is not `kNoCompression`. We should ignore `blob_file_compression_options .enabled`, and mention it here.

src/blob_file_builder.cc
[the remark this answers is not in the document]
yiwu-arbug: Add the following comment before the previous line?
```
// Encode to take ownership of underlying string.
```

src/blob_file_builder.cc
yiwu-arbug: Instead of callback, how about just make caller of `Add()` pass in a `vector<BlobHandle*>`, and batch return the handles when the builder exit buffer state?

Better still, instead of vector you can use `autovector` in rocksdb code, which optimize for small vector.

include/titan/options.h
chux0519: I will add more comments for this, and check if it handles this situation correctly

src/blob_file_builder.cc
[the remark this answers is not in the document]
chux0519: @Connor1996 I think it's OK to use string, and I will add the comment mentioned by @yiwu-arbug, it will be more clear

src/blob_file_builder.cc
```diff
@@ -39,11 +59,81 @@ void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
...
+    Add(rec, &handle);
+  }
+  sample_records_.clear();
+  sample_str_len_ = 0;
+}
+
+void BlobFileBuilder::WriteRawBlock(const Slice& block, BlockHandle* handle) {
+  handle->set_offset(file_->GetFileSize());
+  handle->set_size(block.size());
+  status_ = file_->Append(block);
```
yiwu-arbug, line 108: Let's follow rocksdb's block based table format, add 1 byte compression type and 4 bytes checksum to the end of each block. For compression type, we can use `kNoCompression` for both of the index block and the dictionary block, and leave the compression type byte as a placeholder. For the checksum let's just do crc32.

https://github.com/facebook/rocksdb/blob/master/table/block_based/block_based_table_builder.cc#L1054

src/blob_file_builder.cc
```diff
@@ -6,18 +6,38 @@ namespace titandb {
...
+    header.flags |= BlobFileHeader::kHasUncompressionDictionary;
   std::string buffer;
   header.EncodeTo(&buffer);
   status_ = file_->Append(buffer);
 }
 
 void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
   if (!ok()) return;
+  if (builder_state_ == BuilderState::kBuffered) {
+    std::string rec_str;
```
yiwu-arbug, line 27: nit: s/rec_str/record_str/

let's use abbreviation less for variables. same below.

src/blob_file_builder.cc
```diff
@@ -39,11 +59,81 @@ void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
   }
 }
 
+void BlobFileBuilder::EnterUnbuffered() {
+  // Using collected samples to train the compression dictionary
+  // Then replay those records in memory, encode them to blob file
+  // When above things are done, transform builder state into unbuffered
+  std::string samples = "";
```
yiwu-arbug, line 67:
```
samples.reserve(sample_str_len_);
```

src/blob_file_builder.cc
```diff
@@ -39,11 +59,81 @@ void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
...
+
+  builder_state_ = BuilderState::kUnbuffered;
+
+  // add history buffer
+  BlobHandle handle;
+  for (const std::string& rec_str : sample_records_) {
+    BlobRecord rec;
+    Slice rec_slice(rec_str);
+    rec.DecodeFrom(&rec_slice);
+    Add(rec, &handle);
```
yiwu-arbug, line 99: recursive calling `Add()` is convoluted. how about extract the file write logic into a method, and call it instead.

src/blob_file_builder.cc
```diff
@@ -39,11 +59,81 @@ void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
...
+  encoder_.SetCompressionDict(*compression_dict_);
+
+  builder_state_ = BuilderState::kUnbuffered;
+
+  // add history buffer
+  BlobHandle handle;
+  for (const std::string& rec_str : sample_records_) {
+    BlobRecord rec;
+    Slice rec_slice(rec_str);
+    rec.DecodeFrom(&rec_slice);
```
yiwu-arbug, line 98: You probably don't need to decode the bytes. With some refactoring you can compress the encoded bytes instead.

src/blob_file_builder.cc
chux0519: @yiwu-arbug sounds much simpler, I will try this way

yiwu-arbug: > I think we can buffer the key first, and return the blob handle by `BlobFileBuidler->Finish` then write the key and handle at that time. What do you think @yiwu-arbug

ah, didn't see this comment. I think this is a better idea. @chux0519

chux0519: I guess add another param(vector<BlobHandle*>) would be easier. In this way, we only need to check if the vector is empty or not before using handles we cant. Move the write logic to `Finish` sounds like I will refactor the caller function logic.
Still not sure, I will look into related code closer later.

src/blob_file_builder.cc
chux0519: sure, I've added them after meta index block and dictionary block.

src/blob_file_builder.cc
yiwu-arbug: You do need to refactor the caller, because there's still chance `EnterUnbuffer()` is done on `Finish()`.

src/blob_format.h
```diff
@@ -61,15 +61,30 @@ struct BlobRecord {
...
+                    CompressionDict::GetEmptyDict()) {}
+  BlobEncoder(CompressionType compression,
+              const CompressionDict& compression_dict)
+      : BlobEncoder(compression, CompressionOptions(), compression_dict) {}
+  BlobEncoder(CompressionType compression, CompressionOptions compression_opt)
+      : BlobEncoder(compression, compression_opt,
+                    CompressionDict::GetEmptyDict()) {}
 
   void EncodeRecord(const BlobRecord& record);
+  void EncodeString(const std::string& record_str);
```
yiwu-arbug, line 82: let's do `void EncodeString(const Slice& record)` instead.

src/blob_file_builder.cc
```diff
@@ -27,23 +107,56 @@ void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
...
+    // follow rocksdb's block based table format
+    char trailer[kBlockTrailerSize];
+    // only compression dictionary and meta index block are written
+    // by this method, we use `kNoCompression` as placeholder
+    trailer[0] = kNoCompression;
+    char* trailer_without_type = trailer + 1;
+
+    // crc32 checksum
+    auto crc = crc32c::Value(block.data(), block.size());
+    crc = crc32c::Extend(crc, trailer, 1);  // Extend to cover block type
```
yiwu-arbug, line 127:
```suggestion
    crc = crc32c::Extend(crc, trailer, 1);  // Extend to cover compression type
```

yiwu-arbug (commented):
Looking good overall. The only main issue is handling the blob handle.

src/blob_format.cc
```diff
@@ -40,11 +40,11 @@ void BlobEncoder::EncodeRecord(const BlobRecord& record) {
   EncodeString(record_buffer_);
 }
 
-void BlobEncoder::EncodeString(const std::string& record_str) {
+void BlobEncoder::EncodeString(const Slice& record) {
```
yiwu-arbug, line 43: Change the name to `EncodeSlice`

yiwu-arbug (commented):
only briefly read but looking good overall. Some refactor comments.

src/blob_file_builder.h
```diff
@@ -30,24 +32,41 @@ namespace titandb {
...
   BlobFileBuilder(const TitanDBOptions& db_options,
                   const TitanCFOptions& cf_options, WritableFileWriter* file);
 
-  // Adds the record to the file and points the handle to it.
-  void Add(const BlobRecord& record, BlobHandle* handle);
+  // Adds the record to the file, return `BlobIndices`
+  // Notice: return value might be empty when builder is in `kBuffered` state,
+  // and the ctx.index parameter should set its `file_number` before passed in,
+  // the builder will only change the `blob_handle` of it
+  BlobRecordContexts Add(const BlobRecord& record,
```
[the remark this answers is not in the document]
chux0519: @yiwu-arbug I've resolved all conversations, and need another round review
There is still one remaining problem here,and I think I need your help. Travis keep failing on `titan_table_builder_test`, it says

```bash
titan_table_builder_test: /home/travis/build/tikv/titan/rocksdb/util/compression.h:1380: std::__cxx11::string rocksdb::ZSTD_TrainDictionary(const string&, const std::vector<long unsigned int>&, size_t): Assertion `false' failed.
```
seems like a ZSTD stuff and I can't reproduce this on my machine.

check: https://travis-ci.org/github/tikv/titan/jobs/733970921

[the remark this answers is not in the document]
yiwu-arbug: https://github.com/tikv/rocksdb/blob/6.4.tikv/util/compression.h#L1380

It is failing this assert. Looks like the zstd version in travis is too low.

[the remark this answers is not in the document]
yiwu-arbug: I updated travis script for you. Let's see how it goes.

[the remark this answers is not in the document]
yiwu-arbug: I need to update the build script to include compression tests. I'll see if I get time tomorrow. Meanwhile you can add macro checks around dictionary compression code and tests, just like https://github.com/tikv/rocksdb/blob/6.4.tikv/util/compression.h#L1364

chux0519:
No problem,I will do this later

获取 Outlook for iOS<https://aka.ms/o0ukef>
________________________________
发件人: yiwu-arbug <notifications@github.com>
发送时间: Friday, October 9, 2020 12:48:18 PM
收件人: tikv/titan <titan@noreply.github.com>
抄送: Yongsheng Xu <chuxdesign@hotmail.com>; Mention <mention@noreply.github.com>
主题: Re: [tikv/titan] Add dictionary compression support for blob file (#189)


@yiwu-arbug commented on this pull request.

________________________________

In src/blob_file_builder.h<https://github.com/tikv/titan/pull/189#discussion_r502187174>:

>    // Constructs a builder that will store the contents of the file it
   // is building in "*file". Does not close the file. It is up to the
   // caller to sync and close the file after calling Finish().
   BlobFileBuilder(const TitanDBOptions& db_options,
                   const TitanCFOptions& cf_options, WritableFileWriter* file);

-  // Adds the record to the file and points the handle to it.
-  void Add(const BlobRecord& record, BlobHandle* handle);
+  // Adds the record to the file, return `BlobIndices`
+  // Notice: return value might be empty when builder is in `kBuffered` state,
+  // and the ctx.index parameter should set its `file_number` before passed in,
+  // the builder will only change the `blob_handle` of it
+  BlobRecordContexts Add(const BlobRecord& record,


I need to update the build script to include compression tests. I'll see if I get time tomorrow. Meanwhile you can add macro checks around dictionary compression code and tests, just like https://github.com/tikv/rocksdb/blob/6.4.tikv/util/compression.h#L1364

―

src/blob_gc_job.cc
```diff
@@ -281,6 +265,48 @@ Status BlobGCJob::DoRunGC() {
   return s;
 }
 
+void BlobGCJob::BatchWriteNewIndices(BlobFileBuilder::OutContexts& contexts,
```
Connor1996, line 268:
```suggestion
Status BlobGCJob::BatchWriteNewIndices(BlobFileBuilder::OutContexts& contexts,
```

src/blob_file_builder.h
```diff
@@ -30,24 +32,42 @@ namespace titandb {
...
+
   // Constructs a builder that will store the contents of the file it
   // is building in "*file". Does not close the file. It is up to the
   // caller to sync and close the file after calling Finish().
   BlobFileBuilder(const TitanDBOptions& db_options,
                   const TitanCFOptions& cf_options, WritableFileWriter* file);
 
-  // Adds the record to the file and points the handle to it.
-  void Add(const BlobRecord& record, BlobHandle* handle);
+  // Adds the record to the file, return `BlobIndices`
```
Connor1996, line 51: the comment is not consistent with signature

src/table_builder.cc
```diff
@@ -55,16 +56,17 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
              cf_options_.blob_run_mode == TitanBlobRunMode::kNormal) {
     // we write to blob file and insert index
-    std::string index_value;
-    AddBlob(ikey.user_key, value, &index_value);
+    BlobRecord record;
+    record.key = ikey.user_key;
+    record.value = value;
+    BlobFileBuilder::OutContexts contexts;
+    AddBlob(record, ikey, &contexts);
+    AddToBaseTable(contexts);
```
Connor1996, line 64: if it's in buffer state, the blob record wouldn't be written into SST now. Whereas, for the next key-value which is of no need to write into Titan, it will be written into SST immediately. At last, the buffered blob records will be written into SST at `Finish()`. In this case, the SST is out of order.

src/blob_file_builder.cc
```diff
@@ -6,19 +6,144 @@ namespace titandb {
...
+         0);
+  assert(cf_options_.comparator->Compare(record.key, Slice(largest_key_)) >= 0);
+  largest_key_.assign(record.key.data(), record.key.size());
+}
+
+void BlobFileBuilder::EnterUnbuffered(OutContexts* out_ctx) {
+  // Using collected samples to train the compression dictionary
+  // Then replay those records in memory, encode them to blob file
+  // When above things are done, transform builder state into unbuffered
+  std::string samples = "";
```
tabokie, line 87: no need for assignment.

src/table_builder.cc
tabokie: might need to copy the value twice to cachedCtx and outCtx to solve this.

src/blob_gc_job.h
```diff
@@ -90,9 +90,10 @@ class BlobGCJob {
   uint64_t io_bytes_written_ = 0;
 
   Status DoRunGC();
-  Status BuildIterator(std::unique_ptr<BlobFileMergeIterator>* result);
-  Status DiscardEntry(const Slice& key, const BlobIndex& blob_index,
-                      bool* discardable);
+  void BatchWriteNewIndices(BlobFileBuilder::OutContexts &contexts, Status *s);
+  Status BuildIterator(std::unique_ptr<BlobFileMergeIterator> *result);
+  Status DiscardEntry(const Slice &key, const BlobIndex &blob_index,
+                      bool *discardable);
```
tabokie, line 96: format

tabokie (commented):
some minor issues

src/table_builder.cc
tabokie: might need to copy the value twice to out-context and cached-context to solve this.

Connor1996: I think we should introduce buffer state at table builder level.

Connor1996 (commented):
Please add a test case to cover the disorder issue.

src/table_builder.cc
```diff
@@ -52,19 +53,19 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
       base_builder_->Add(key, value);
     }
   } else if (ikey.type == kTypeValue &&
-             value.size() >= cf_options_.min_blob_size &&
```
Connor1996, line 55: How about checking the state of BlobFileBuilder and whether is small kv in `AddToBaseTable` instead of in `AddBlob`. If in buffer state, you can append the small kv into `BlobFileBuilder` by another interface.

If that, you can remove the `HasFileWriter` of `BlobFileBuilder` and still keep the `value.size() >= cf_options_.min_blob_size` logic into `TableBuilder` which makes the logic tight and clear.

chux0519: Yeah, it would be clearer to check `BlobFileBuilder`'s state here. 
But still, the `HasFileWriter` is hard to remove, it is used to allow empty writer when `BlobFileBuilder` construct. Otherwise, we will have a needless blob file with no record when inserting small kv pairs with dictionary compression enabled.
check discussion here: https://tikv-wg.slack.com/archives/CQ545HDGC/p1602417406036300?thread_ts=1602417359.036200&cid=CQ545HDGC

src/table_builder.cc
chux0519: resolved in this commit:  [9073ecca81164f89f36c412e44eac197b80f7197](https://github.com/tikv/titan/pull/189/commits/9073ecca81164f89f36c412e44eac197b80f7197)

discussed here: https://tikv-wg.slack.com/archives/CQ545HDGC/p1602303591032200?thread_ts=1601073597.011400&cid=CQ545HDGC

src/table_builder.cc
```diff
@@ -110,66 +110,119 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+  // Init blob_builder_ first
+  if (!blob_builder_)
+    blob_builder_.reset(new BlobFileBuilder(db_options_, cf_options_, nullptr));
+  assert(blob_builder_);
+
+  bool is_small_kv = ikey.type == kTypeValue &&
+                     record.value.size() < cf_options_.min_blob_size;
+
+  // Only if this record is a small KV pair, can the blob_builder_ has no
+  // writer, otherwise should setup a writer
```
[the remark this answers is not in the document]
chux0519: updated

src/blob_file_iterator_test.cc
```diff
@@ -2,13 +2,12 @@
...
-#include "file/filename.h"
-#include "test_util/testharness.h"
-#include "util/random.h"
-
 #include "blob_file_builder.h"
 #include "blob_file_cache.h"
 #include "blob_file_reader.h"
+#include "file/filename.h"
+#include "test_util/testharness.h"
+#include "util/random.h"
```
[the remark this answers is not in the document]
chux0519: I'm not sure how to fix this, can you give me more hints.

src/blob_gc_job.h
chux0519: I've run `bash ./scripts/format-diff.sh` seems nothing happened. Need more hints.

src/blob_file_iterator_test.cc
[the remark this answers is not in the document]
tabokie: It's simple, just revert the changes that merge multiple include chunks into one. Here just move these three up front like the way it was.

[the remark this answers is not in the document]
tabokie: Also don't forget other files, basically split them into std header, rocksdb header, titan header.

src/blob_gc_job.h
tabokie: Not sure where went wrong. but `T &` and `T *` is clearly not in line with other functions.

src/blob_file_iterator_test.cc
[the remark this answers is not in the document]
chux0519: It seems that `clang-format`'s default config cause this problem. I've tried to add some rules in my `.clang-format` and it works. Check: https://github.com/chux0519/titan/compare/pcp-23-2...pcp-23-2-format

But these rule will fail the formatter test( `git diff "git merge-base master HEAD" | clang-format-diff -style=google -p1 -i`). The test script is unable to set `IncludeCategories` rules, so I'm not sure if I should fix it in this PR,I think the format problem is worth for another issue, we can add a full `.clang-format` for this project, like

```
BasedOnStyle: google

IncludeCategories:
  - Regex:           '^"(titan)/'
    Priority:        4
    SortPriority:    3
  - Regex:           '^"[[:alnum:].|_]+/'
    Priority:        2
    SortPriority:    2
  - Regex:           '<[[:alnum:].|_|/]+>'
    Priority:        1
    SortPriority:    0
  - Regex:           '.*'
    Priority:        4
    SortPriority:    4
```

and change the formatter test script (now using clang-format-diff is unable to set `IncludeCategories` rules), And I'm willing to help with this. @tabokie @Connor1996 @yiwu-arbug  What do you guys think?

src/table_builder.cc
Connor1996: Check `BlobFileBuilder's state here, then there is no need to call `AddBlob` for small kv pairs so that there is no case for needless blob file created.

chux0519: > Check `BlobFileBuilder's state here, then there is no need to call `AddBlob` for small kv pairs so that there is no case for needless blob file created.

yeah, I see. I will refactor this later.

chux0519: updated

src/table_builder.cc
```diff
@@ -52,18 +53,36 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+      record.value = value;
+      BlobFileBuilder::OutContexts contexts;
+      AddBlob(record, ikey, &contexts);
+      AddToBaseTable(contexts);
+      UpdateIOBytes(prev_bytes_read, prev_bytes_written, &io_bytes_read_,
+                    &io_bytes_written_);
+
+      if (ok()) return;
     }
   } else if (ikey.type == kTypeBlobIndex && cf_options_.level_merge &&
```
Connor1996, line 87: For all `base_builder_->Add`, you should check the state of the blob builder. It still suffers from the disorder issue in the case that it is `kTypeBlobIndex` but doesn't hit this if branch.

Connor1996: Please update the disorder test case too

Connor1996 (commented):
Most LGTM, you can prepare to set up a cluster to test it.

chux0519:
> Most LGTM, you can prepare to set up a cluster to test it.

Nice! But I did not implement the decode logic yet, should I do that first ?

I think for now, I will

1. fix the remaining disorder issue (the `kTypeBlobIndex` case)
2. implement decompress logic

At the same time I can prepare the cluster setup, can you provide me more details or references, I think I need some learning materials about the cluster testing and benchmarking stuff.

Connor1996:
> > Most LGTM, you can prepare to set up a cluster to test it.
> 
> Nice! But I did not implement the decode logic yet, should I do that first ?
> 
> I think for now, I will
> 
> 1. fix the remaining disorder issue (the `kTypeBlobIndex` case)
> 2. implement decompress logic
> 
> At the same time I can prepare the cluster setup, can you provide me more details or references, I think I need some learning materials about the cluster testing and benchmarking stuff.

Oh, forget it. It's still some works to do before it can be enabled. After finishing decompression, you can use `titandb_bench` to run some workloads. Then, you can try to deploy a cluster with Titan enabled. See:
- https://docs.pingcap.com/zh/tidb/stable/production-deployment-using-tiup
- https://docs.pingcap.com/zh/tidb/stable/titan-configuration

chux0519:
@Connor1996 Thanks! I will check it out.

src/blob_file_iterator_test.cc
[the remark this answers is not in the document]
yiwu-arbug: Let's follow default clang-format formatter and don't add custom styling rules.

src/table_builder.cc
```diff
@@ -87,35 +116,43 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+    }
   } else {
+    assert(builder_unbuffered());
     base_builder_->Add(key, value);
   }
 }
 
-void TitanTableBuilder::AddBlob(const Slice& key, const Slice& value,
-                                std::string* index_value) {
+void TitanTableBuilder::AddBlob(const BlobRecord& record,
```
yiwu-arbug, line 148:
```
void TitanTableBuilder::AddBlob(
    const Slice& key, const Slice& value,
    const ParsedInternalKey& ikey,
    BlobFileBuilder::OutContexts* contexts) {
```
and construct `BlobRecord` inside the method?

src/table_builder.cc
```diff
@@ -126,36 +163,63 @@ void TitanTableBuilder::AddBlob(const Slice& key, const Slice& value,
         new BlobFileBuilder(db_options_, cf_options_, blob_handle_->GetFile()));
   }
 
-  RecordTick(statistics(stats_), TITAN_BLOB_FILE_NUM_KEYS_WRITTEN);
```
yiwu-arbug, line 129: should keep the TITAN_BLOB_FILE_NUM_KEYS_WRITTEN counter

src/blob_file_builder.cc
```diff
@@ -6,19 +6,132 @@ namespace titandb {
...
+
+  FlushSampleRecords(out_ctx);
+
+  builder_state_ = BuilderState::kUnbuffered;
+}
+
+void BlobFileBuilder::FlushSampleRecords(OutContexts* out_ctx) {
+  assert(cached_contexts_.size() >= sample_records_.size());
+  size_t i = 0, j = 0;
+  for (; i < sample_records_.size(); i++, j++) {
```
yiwu-arbug, line 115: what does this double loop do? it doesn't look correct.

src/table_builder.cc
```diff
@@ -87,35 +116,43 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
       // doing level merge.
       if (get_status.ok()) {
         std::string index_value;
-        AddBlob(ikey.user_key, record.value, &index_value);
+        BlobRecord blob_record;
+        blob_record.key = ikey.user_key;
+        blob_record.value = record.value;
+        BlobFileBuilder::OutContexts contexts;
+        AddBlob(blob_record, ikey, &contexts);
+        AddToBaseTable(contexts);
```
yiwu-arbug, line 124: Since we call `AddToBaseTable` every time after `AddBlob`, let's move the `AddToBaseTable` call to inside `AddBlob`.

src/table_builder.cc
```diff
@@ -126,36 +163,63 @@ void TitanTableBuilder::AddBlob(const Slice& key, const Slice& value,
...
+        ctx->new_blob_index.EncodeTo(&index_value);
+
+        ikey.type = kTypeBlobIndex;
+        std::string index_key;
+        AppendInternalKey(&index_key, ikey);
+        base_builder_->Add(index_key, index_value);
+      }
     }
   }
+  if (!finishing && blob_handle_->GetFile()->GetFileSize() >=
```
yiwu-arbug, line 207: We should have this check and call `FinishBlobFile` inside `AddBlob`, before calling `AddToBaseTable`.

yiwu-arbug:
Almost there! Only a bunch of minor comments, mostly for code cleanup. And I'm working on #191 to enable compression tests in travis, but need to fix some tests.

src/blob_file_builder.cc
chux0519: to skip the "cached" contexts, small KV pairs or blobindex during merge process would have chance in here. they should be returned the same as they were passed in. after the inner loop, the context should match the sample record.

yiwu-arbug:
@chux0519 you can rebase to include #191, which enable zstd in travis.

src/blob_file_builder.cc
yiwu-arbug: give `i` and `j` a better name? like `sample_idx` and `ctx_idx`.

src/blob_file_builder.cc
```diff
@@ -6,19 +6,132 @@ namespace titandb {
...
-      encoder_(cf_options_.blob_file_compression) {
+      encoder_(cf_options.blob_file_compression,
+               cf_options.blob_file_compression_options) {
+#if ZSTD_VERSION_NUMBER < 10103
+  if (cf_options_.blob_file_compression_options.max_dict_bytes > 0) {
+    status_ = Status::NotSupported("ZSTD version too old.");
+    return;
+  }
+#endif
+  if (file_) WriteHeader();
```
yiwu-arbug, line 22: can you explain how `file_` is handle when it is not initially set?

chux0519:
> @chux0519 you can rebase to include #191, which enable zstd in travis.

sure I will do this.

src/blob_file_builder.cc
chux0519: this is not needed anymore, I'll update here.

src/table_builder.cc
chux0519: I found out actually 

```
void TitanTableBuilder::AddBlob(const ParsedInternalKey& ikey,
                                const Slice& value,
                                BlobFileBuilder::OutContexts* contexts)
```
is enough, key can be read from `ikey.user_key`

src/blob_gc_job.cc
```diff
@@ -379,7 +405,9 @@ Status BlobGCJob::Finish() {
 Status BlobGCJob::InstallOutputBlobFiles() {
   Status s;
   for (auto& builder : blob_file_builders_) {
-    s = builder.second->Finish();
+    BlobFileBuilder::OutContexts contexts;
+    s = builder.second->Finish(&contexts);
```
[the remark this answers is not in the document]
chux0519: Actually I am a little bit confused of  `InstallOutputBlobFiles` , can you tell me more about this function, and give some hints(ex. when and where we will switch a file, and which are the key functions), at the same time, I will take a closer look into this.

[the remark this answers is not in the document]
yiwu-arbug: The blob GC job rewrites blob file into new blob files and discard deleted data on the way. The new blob files are not part of the DB (e.g. if DB crash in the middle of a GC job, those new files will be discarded after restart) until `InstallOutputBlobFiles` is called. `InstallOutputBlobFiles` write the list of new blob files to the DB's manifest. After that the new files are considered part of the DB.

src/blob_format.h
```diff
@@ -61,15 +61,30 @@ struct BlobRecord {
...
  public:
-  BlobEncoder(CompressionType compression,
+  BlobEncoder(CompressionType compression, CompressionOptions compression_opt,
               const CompressionDict& compression_dict)
-      : compression_ctx_(compression),
-        compression_info_(compression_opt_, compression_ctx_, compression_dict,
-                          compression, 0 /*sample_for_compression*/) {}
+      : compression_opt_(compression_opt),
+        compression_ctx_(compression),
+        compression_info_(new CompressionInfo(compression_opt, compression_ctx_,
```
yiwu-arbug, line 68:
```suggestion
        compression_info_(new CompressionInfo(compression_opt_, compression_ctx_,
```
Also the compression_dict need to be a member of `BlobEncoder`. This will fix CI.

yiwu-arbug:
mostly good! looking forward to merge this one.

src/blob_format.h
chux0519: I will try this to fix this later.

I can't reproduce the stack-buffer-underflow with gcc or clang-10 (both with asan). It's a little bit strange. Also when using clang-10, it will throw a rocksdb part error, saying 
```
In file included from /home/hexyoungs/workspace/pingcap/rocksdb/utilities/backupable/backupable_db.cc:19:
/home/hexyoungs/workspace/pingcap/rocksdb/util/channel.h:35:33: error: no matching constructor for initialization of 'std::lock_guard<std::mutex>'
    std::lock_guard<std::mutex> lk(lock_);
```

with following patch(tikv/rocksdb) it will compile, but still will not hit the `stack-buffer-underflow` error in test cases.

```
diff --git a/util/channel.h b/util/channel.h
index 0225482c0..2eaf02d0c 100644
--- a/util/channel.h
+++ b/util/channel.h
@@ -31,7 +31,7 @@ class channel {
     return buffer_.empty() && eof_;
   }

-  size_t size() const {
+  size_t size() {
     std::lock_guard<std::mutex> lk(lock_);
     return buffer_.size();
   }
```

src/blob_gc_job.cc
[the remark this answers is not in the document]
chux0519: yeah, I see. so this loop here is not necessary, I will move the `Finish` call down to the main process loop, right before file switching

src/blob_format.h
chux0519: reproduced in a VM, ubuntu20 with clang-7, I'll figure out what's happening and fix this tomorrow.

yiwu-arbug: trust me, just apply the fix I said..

src/blob_gc_job.cc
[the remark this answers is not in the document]
yiwu-arbug: looks like this change is not done yet?

src/blob_format.h
chux0519: sure, I'll try

src/blob_gc_job.cc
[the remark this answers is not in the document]
chux0519: @yiwu-arbug I removed the first loop (which is just finishing all blob files), and move the single finish action in the following loop. Is this correct?

before: 
> loop {finish all files} -> loop {files.emplace_back(new blob)}

after: 
> loop {finish current file -> files.emplace_back(new blob) }

src/blob_format.h
chux0519: I don't quite understand this, but it works, awesome!

yiwu-arbug: params of CompressionInfo constructor are references, not passing by value.

yiwu-arbug (commented):
LGTM. defer to @Connor1996 for final review and merge.

src/blob_format.h
chux0519: yeah, I got it. The problem is  about the first param. And after playing this around, actually change `compression_opt` to `compression_opt_` is enough to fix the ASAN error. But yeah we better keep `compression_dict` as well.

src/blob_file_builder.cc
```diff
@@ -27,23 +128,56 @@ void BlobFileBuilder::Add(const BlobRecord& record, BlobHandle* handle) {
...
-  std::string buffer;
+  if (builder_state_ == BuilderState::kBuffered) {
+    EnterUnbuffered(out_ctx);
+  }
+
   BlobFileFooter footer;
+  // if has compression dictionary, encode it into meta blocks
+  // and update relative fields in footer
+  if (cf_options_.blob_file_compression_options.max_dict_bytes > 0) {
+    BlockHandle meta_index_handle, uncompression_dict_handle;
```
Connor1996, line 173: `uncompression_dict_handle` is useless. As commented before, no need to set `uncompression_dict_handle` to `footer`. Seems we should remove the field from `BlobFileFooter`. /cc @yiwu-arbug

Connor1996 (commented):
seems the test doesn't cover the disorder case that level-merge enabled.

chux0519:
> seems the test doesn't cover the disorder case that level-merge enabled.

ah, I'll add this, and also remove the `uncompression_dict_handle` in footer

src/blob_file_builder.cc
chux0519: actually `uncompression_dict_handle` is removed before, and since it's useless here, I refactored `WriteCompressionDictBlock`, to allocate a block handle inside the method.

yiwu-arbug:
Awesome!

PR #192, opened 2020-10-29
implement dictionary compression decompression

chux0519:
This is a subtask of issue: https://github.com/tikv/tikv/issues/5743

> Read blob from file with dictionary compression

Mainly made following changes:

- Refactored `BlobDecoder`, add decompression related components in it.
- Refactored `BlobFileReader` and `BlobFileIterator`, initialize the decoder with corresponding data in blob files.
- add testcases

Issue Number: https://github.com/tikv/tikv/issues/8635


src/blob_format.cc
```diff
@@ -364,5 +363,79 @@ bool operator==(const BlobFileFooter& lhs, const BlobFileFooter& rhs) {
...
+  }
+  BlockContents meta_block_content(std::move(ubuf), meta_index_handle.size());
+
+  std::unique_ptr<Block> meta(
+      new Block(std::move(meta_block_content), kDisableGlobalSequenceNumber));
+
+  std::unique_ptr<InternalIterator> meta_iter(
+      meta.get()->NewDataIterator(BytewiseComparator(), BytewiseComparator()));
+
+  bool dict_is_found = false;
```
Connor1996, line 418: seems it's not used. we should return an error.

src/table_builder_test.cc
```diff
@@ -400,13 +400,32 @@ TEST_F(TableBuilderTest, DictCompress) {
...
+    ASSERT_OK(DecodeInto(iter->value(), &index));
+    ASSERT_EQ(index.file_number, kTestFileNumber);
+    BlobRecord record;
+    PinnableSlice buffer;
+    ASSERT_OK(blob_reader->Get(ro, index.blob_handle, &record, &buffer));
+    ASSERT_EQ(record.key, key);
+    ASSERT_EQ(record.value, std::string(kMinBlobSize, i));
+    iter->Next();
+  }
+
```
tabokie, line 428: check iter invalid

src/blob_format.cc
```diff
@@ -364,5 +363,79 @@ bool operator==(const BlobFileFooter& lhs, const BlobFileFooter& rhs) {
...
 
+// Seek to the specified meta block.
+// Return true if it successfully seeks to that block.
+static Status SeekToMetaBlock(InternalIterator* meta_iter,
+                              const std::string& block_name, bool* is_found,
+                              BlockHandle* block_handle = nullptr) {
+  if (block_handle != nullptr) {
+    *block_handle = BlockHandle::NullBlockHandle();
+  }
+  *is_found = true;
```
tabokie, line 374: should be false

src/blob_format.cc
```diff
@@ -364,5 +363,79 @@ bool operator==(const BlobFileFooter& lhs, const BlobFileFooter& rhs) {
...
+    if (meta_iter->Valid() && meta_iter->key() == block_name) {
+      *is_found = true;
+      if (block_handle) {
+        Slice v = meta_iter->value();
+        return block_handle->DecodeFrom(&v);
+      }
+    } else {
+      *is_found = false;
+      return Status::OK();
+    }
```
tabokie, line 386: this branch can be removed

src/table_builder_test.cc
```diff
@@ -400,13 +400,32 @@ TEST_F(TableBuilderTest, DictCompress) {
...
 #endif
 }
```
tabokie, line 430: maybe add a test where a vanilla decoder tries to decode compressed block?

chux0519: Do you mean use a BlobFileReader with the default TitanCFOptions, to decode the compressed block?

tabokie: Yes, to check the exception safety for possible misconfig.

src/blob_file_iterator.cc
```diff
@@ -48,7 +44,26 @@ bool BlobFileIterator::Init() {
...
   status_ = blob_file_footer.DecodeFrom(&slice);
   end_of_blob_record_ = file_size_ - BlobFileFooter::kEncodedLength -
-                        blob_file_footer.meta_index_handle.size();
+                        blob_file_footer.meta_index_handle.size() -
+                        kBlockTrailerSize;
+
+  if (blob_file_header.flags & BlobFileHeader::kHasUncompressionDictionary) {
+    status_ = InitUncompressionDict(blob_file_footer, file_.get(),
+                                    &uncompression_dict_);
+    decoder_->SetUncompressionDict(uncompression_dict_.get());
```
tabokie, line 53: We should exit before using it when `!status.ok()`

src/blob_format.cc
```diff
@@ -364,5 +365,78 @@ bool operator==(const BlobFileFooter& lhs, const BlobFileFooter& rhs) {
...
+  s = SeekToMetaBlock(meta_iter.get(), kCompressionDictBlock, &dict_is_found,
+                      &dict_block);
+
+  if (!dict_is_found) {
+    return Status::NotFound("uncompression dict");
+  }
+
+  if (!s.ok()) {
+    return s;
+  }
```
tabokie, line 426: swap places

src/blob_file_iterator.h
```diff
@@ -58,7 +58,10 @@ class BlobFileIterator {
   Status status_;
   bool valid_{false};
 
-  BlobDecoder decoder_;
+  std::unique_ptr<UncompressionDict> uncompression_dict_;
+  std::unique_ptr<BlobDecoder> decoder_ =
+      std::unique_ptr<BlobDecoder>(new BlobDecoder);
```
tabokie, line 63: remind me why we need to make this a pointer?

chux0519: The decoder has to reconstruct at my first implementaion, so it needed a smart pointer. But, I've done some refactoring in last few commits, the decoder here doesn't really need a smart pointer, I'll change it! Nice catch!

src/blob_file_iterator.cc
```diff
@@ -48,7 +44,26 @@ bool BlobFileIterator::Init() {
   BlobFileFooter blob_file_footer;
   status_ = blob_file_footer.DecodeFrom(&slice);
   end_of_blob_record_ = file_size_ - BlobFileFooter::kEncodedLength -
-                        blob_file_footer.meta_index_handle.size();
+                        blob_file_footer.meta_index_handle.size() -
+                        kBlockTrailerSize;
```
yiwu-arbug, line 48: It seems if meta_index_handle is empty, we should not subtract `kBlockTrailerSize`?

yiwu-arbug (commented):
Looking good overall.

src/blob_file_iterator.cc
chux0519: yeah, will fix it

src/blob_format.h
```diff
@@ -388,5 +396,12 @@ Status DecodeInto(const Slice& src, T* target) {
   return s;
 }
 
+// Init uncompression dictionary
+// called by BlobFileReader and BlobFileIterator when blob file has
+// uncompression dictionary
+Status InitUncompressionDict(
```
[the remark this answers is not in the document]
chux0519: it also called from blob_file_iterator.cc, I think it would be better to leave it here.

src/blob_format.cc
```diff
@@ -364,5 +365,78 @@ bool operator==(const BlobFileFooter& lhs, const BlobFileFooter& rhs) {
           lhs.meta_index_handle.size() == rhs.meta_index_handle.size());
 }
 
+// Seek to the specified meta block.
+// Return true if it successfully seeks to that block.
+static Status SeekToMetaBlock(InternalIterator* meta_iter,
```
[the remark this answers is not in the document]
chux0519: I've learned. will refactor this.

src/blob_format.h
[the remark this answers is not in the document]
yiwu-arbug: have blob_file_iterator.cc include blob_file_reader.h ?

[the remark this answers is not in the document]
chux0519: no problem, will do this

src/table_builder_test.cc
```diff
@@ -400,13 +400,98 @@ TEST_F(TableBuilderTest, DictCompress) {
...
+    std::string value;
+    value = std::string(kMinBlobSize, i);
+    table_builder->Add(ikey.Encode(), value);
+  }
+  ASSERT_OK(table_builder->Finish());
+  ASSERT_OK(base_file->Sync(true));
+  ASSERT_OK(base_file->Close());
+  BlobFileExists(true);
+
+  // reset cf_options_, decode using vanilla reader should also works
```
tabokie, line 463: Why is a plain decoder able to read compressed information? Am I missing something?

chux0519: BlobFileReader will respect the header of blob files (which indicate whether the file has uncompression dictionary or not).
And the decoder is on "record" level, which means, every time we want to decode a record, we will create a decoder using the uncompression dictionary owned by the reader, it firstly decode the record header(which will tell what compression method the record use), then decode the content.so actually in the whole decoding process, the options passed in has not be used, the file reader will detect from the blob file itself.

tabokie: I see that compression type is hardcoded in blob file, so cf_options is not used by blob reader. Do you think we need to error when those two aren't consistent? @yiwu-arbug

yiwu-arbug: I see. It is okay to have inconsistency in the compression options in cf_options and blob file. The cf_options only tells what compression to use for *new* blob files.

yiwu-arbug (approved):
LGTM. Defer to @tabokie for accepting.

src/table_builder_test.cc
```diff
@@ -400,13 +400,98 @@ TEST_F(TableBuilderTest, DictCompress) {
...
+    ASSERT_OK(blob_reader->Get(ro, index.blob_handle, &record, &buffer));
+    ASSERT_EQ(record.key, key);
+    ASSERT_EQ(record.value, std::string(kMinBlobSize, i));
+    iter->Next();
+  }
+  ASSERT_TRUE(!iter->Valid());
+#endif
+}
+
+TEST_F(TableBuilderTest, VanillaDecodeDictCompress) {
```
tabokie, line 432: Now that cf-options isn't picked up by reader, we can remove this test. Sorry for assigning this extra work.

tabokie (approved):
Rest looks good!

ti-srebot (approved):

src/table_builder_test.cc
chux0519: sure no problem

Connor1996 (approved):

PR #202, opened 2021-01-19
fix titan zstd dictionary compression bug

ZhenhanGong:
## Bug 1
When max_dict_bytes is set and max_train_bytes equals 0, data insertion will go wrong. Because **Finish()** will never get invoked since **NumEntries()** returns 0.

### What happened:
When max_dict_bytes is set and max_train_bytes equals 0, blob file will retain **kBuffered** state. 
It counts on BlobFileBuilder::Finish() to invoke **EnterUnbuffered()**, which will invoke WriteEncoderData() to actually write data. 
However, TitanTableBuilder::**NumEntries()** will return 0, and then in rocksdb::builder.cc **empty** will be true, result in **Finsh()** never get invoked, and all data is lost.

### How to fix:
We can fix the bug via return counting sample records num as well and return the number of samples records in TitanTableBuilder::**NumEntries()**. That way **Finish()** will be invoked properly.

## Bug 2
The implementation of FlushSampleRecords() forget to count non-blob kv items, resulting in assertion of ctx_idx == size will fail sometimes. 

### How to fix:
Add a for loop to count remaining non-blob kv items after counting all sample records.

Signed-off-by: Zhenhan Gong <zhenhan.gong@gmail.com>

### DIFF
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -112,9 +112,8 @@ void BlobFileBuilder::FlushSampleRecords(OutContexts* out_ctx) {
     WriteEncoderData(&ctx->new_blob_index.blob_handle);
     out_ctx->emplace_back(std::move(cached_contexts_[ctx_idx]));
   }
-  for (; ctx_idx < cached_contexts_.size() &&
-          cached_contexts_[ctx_idx]->has_value;
-        ctx_idx++) {
+  for (; ctx_idx < cached_contexts_.size(); ctx_idx++) {
+    assert(cached_contexts_[ctx_idx]->has_value);
     out_ctx->emplace_back(std::move(cached_contexts_[ctx_idx]));
   }
   assert(sample_idx == sample_records_.size());
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -112,6 +112,11 @@ void BlobFileBuilder::FlushSampleRecords(OutContexts* out_ctx) {
     WriteEncoderData(&ctx->new_blob_index.blob_handle);
     out_ctx->emplace_back(std::move(cached_contexts_[ctx_idx]));
   }
+  for (; ctx_idx < cached_contexts_.size() &&
+          cached_contexts_[ctx_idx]->has_value;
+        ctx_idx++) {
+    out_ctx->emplace_back(std::move(cached_contexts_[ctx_idx]));
+  }
   assert(sample_idx == sample_records_.size());
   assert(ctx_idx == cached_contexts_.size());
   sample_records_.clear();
--- a/src/table_builder.cc
+++ b/src/table_builder.cc
@@ -290,10 +290,11 @@ void TitanTableBuilder::Abandon() {
 }
 
 uint64_t TitanTableBuilder::NumEntries() const {
-  if (builder_unbuffered())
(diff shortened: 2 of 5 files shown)


yiwu-arbug (commented):
Can we add a unit test?

src/table_builder_test.cc
```diff
@@ -429,6 +429,47 @@ TEST_F(TableBuilderTest, DictCompress) {
 #endif
 }
 
+TEST_F(TableBuilderTest, DictCompressOptions) {
```
Connor1996, line 432: Please write the test in the Titan level, that is to say, new a titan instance and write some by db interface instead of putting into `TitanTableBuilder` directly

src/titan_db_test.cc
```diff
@@ -380,6 +380,27 @@ TEST_F(TitanDBTest, Basic) {
...
+
+  const uint64_t kNumKeys = 501;
+  std::map<std::string, std::string> data;
+  Open();
+  for (uint64_t k = 1; k <= kNumKeys; k++) {
+    Put(k, &data);
+  }
+  Flush();
+  VerifyDB(data);
+  printf("gzh: end of TitanDBTest::DictCompressOptions\n");
```
Connor1996, line 401: please clean this

Connor1996 (approved):

ZhenhanGong:
Add a minor bug fix to correctly count ctx_idx inside FlushSampleRecords().

yiwu-arbug:
please run scripts/format-diff.sh once.

src/table_builder.cc
```diff
@@ -290,7 +290,10 @@ void TitanTableBuilder::Abandon() {
 }
 
 uint64_t TitanTableBuilder::NumEntries() const {
-  return base_builder_->NumEntries();
+  if (builder_unbuffered())
+    return base_builder_->NumEntries();
+  else
+    return blob_builder_->NumEntries() + blob_builder_->NumSampleEntries();
```
yiwu-arbug, line 296:
```suggestion
  if (builder_unbuffered()) {
    return base_builder_->NumEntries();
  } else {
    return blob_builder_->NumEntries() + blob_builder_->NumSampleEntries();
  }
```

src/blob_file_builder.cc
```diff
@@ -112,6 +112,11 @@ void BlobFileBuilder::FlushSampleRecords(OutContexts* out_ctx) {
     WriteEncoderData(&ctx->new_blob_index.blob_handle);
     out_ctx->emplace_back(std::move(cached_contexts_[ctx_idx]));
   }
+  for (; ctx_idx < cached_contexts_.size() &&
+          cached_contexts_[ctx_idx]->has_value;
+        ctx_idx++) {
+    out_ctx->emplace_back(std::move(cached_contexts_[ctx_idx]));
+  }
```
yiwu-arbug, line 119:
```suggestion
  for (; ctx_idx < cached_contexts_.size(); ctx_idx++) {
    assert(cached_contexts_[ctx_idx]->has_value); // or better have some error handling
    out_ctx->emplace_back(std::move(cached_contexts_[ctx_idx]));
  }
```

yiwu-arbug (approved):
rest LGTM

ti-srebot (approved):

PR #207, opened 2021-04-21
Add checkpoint support

caipengbo:
I added checkpoint support for Titan inspired by [tikv/rocksdb](https://github.com/tikv/rocksdb/tree/6.4.tikv/).

The basic idea is:
1. Create a new, separate directory for checkpoint
2. Call `TitanDB::DisableFileDeletions` to disable delete operations
3. Call `rocksdb::CheckpointImpl::CreateCheckpoint` to create base db checkpoint
4. Call `TitanDB::GetAllTitanFiles()` to collect all existing files and record all existing blob file in `titandb::VersionEdit version_edit`
5. For all existing files: 
    - If the Checkpoint directory and the source database directory are on the same file system, hard linked blob files, otherwise copy them
    - For the MANIFEST file, create it based on `version_edit`, which will prevent removing useful blob file when Titan reopen 
    - For the CURRENT file, read the name of the MANIFEST file and write it into the CURRENT file
6. Call `TitanDB::EnableFileDeletions()` to restore deletion

This will include redundant blob files in the checkpoint, but hopefully not a lot of them, Titan will recalculate GC stats and GC out those redundant blob files on restart.

### DIFF
--- a/src/blob_file_set.cc
+++ b/src/blob_file_set.cc
@@ -365,8 +365,8 @@ void BlobFileSet::GetAllFiles(std::vector<std::string>* files,
   }
 
   // Append current MANIFEST and CURRENT file name
-  files->emplace_back(DescriptorFileName("/titandb", manifest_file_number_));
-  files->emplace_back(CurrentFileName("/titandb"));
+  files->emplace_back(DescriptorFileName("", manifest_file_number_));
+  files->emplace_back(CurrentFileName(""));
 }
 
 }  // namespace titandb
--- a/src/blob_file_set.cc
+++ b/src/blob_file_set.cc
@@ -153,13 +153,16 @@ Status BlobFileSet::OpenManifest(uint64_t file_number) {
     ImmutableDBOptions ioptions(db_options_);
     s = SyncTitanManifest(env_, stats_, &ioptions, manifest_->file());
   }
+  uint64_t old_manifest_file_number = manifest_file_number_;
   if (s.ok()) {
     // Makes "CURRENT" file that points to the new manifest file.
     s = SetCurrentFile(env_, dirname_, file_number, nullptr);
+    manifest_file_number_ = file_number;
   }
 
   if (!s.ok()) {
     manifest_.reset();
+    manifest_file_number_ = old_manifest_file_number;
     obsolete_manifests_.emplace_back(file_name);
   }
   return s;
@@ -326,22 +329,16 @@ void BlobFileSet::GetObsoleteFiles(std::vector<std::string>* obsolete_files,
   obsolete_manifests_.clear();
 }
 
-void BlobFileSet::GetLiveFiles(std::vector<std::string>* live_files, uint64_t* manifest_file_size) {
-  for (auto it = column_families_.begin(); it != column_families_.end();) {
-    auto& cf_id = it->first;
-
-    if (obsolete_columns_.find(cf_id) == obsolete_columns_.end()) {
-      auto& blob_storage = it->second;
+void BlobFileSet::GetAllFiles(std::vector<std::string>* files, uint64_t* manifest_file_size) {
+  for (auto it = column_families_.begin(); it != column_families_.end(); ++it) {
+    auto& blob_storage = it->second;
       // get live blob file
-      blob_storage->GetLiveFiles(live_files);
-    }
-
-    ++it;
+    blob_storage->GetAllFiles(files);
   }
   // append current MANIFEST and CURRENT file name
-  std::string file_name = DescriptorFileName("/titandb", next_file_number_.load()-1);
-  live_files->push_back(file_name);
-  live_files->push_back(CurrentFileName("/titandb"));
+  std::string file_name = DescriptorFileName("/titandb", manifest_file_number_);
+  files->push_back(file_name);
+  files->push_back(CurrentFileName("/titandb"));
 
   // find current length of titandb manifest file while holding the mutex lock
   *manifest_file_size = manifest_->file()->GetFileSize();
--- a/src/blob_file_set.cc
+++ b/src/blob_file_set.cc
@@ -329,7 +329,7 @@ void BlobFileSet::GetObsoleteFiles(std::vector<std::string>* obsolete_files,
   obsolete_manifests_.clear();
 }
 
-void BlobFileSet::GetAllFiles(std::vector<std::string>* files, 
+void BlobFileSet::GetAllFiles(std::vector<std::string>* files,
                               std::vector<VersionEdit>* edits) {
   std::vector<std::string> all_blob_files;
 
--- a/src/blob_file_set.cc
+++ b/src/blob_file_set.cc
@@ -329,19 +329,40 @@ void BlobFileSet::GetObsoleteFiles(std::vector<std::string>* obsolete_files,
   obsolete_manifests_.clear();
 }
 
-void BlobFileSet::GetAllFiles(std::vector<std::string>* files, uint64_t* manifest_file_size) {
-  for (auto it = column_families_.begin(); it != column_families_.end(); ++it) {
-    auto& blob_storage = it->second;
-      // get live blob file
-    blob_storage->GetAllFiles(files);
-  }
-  // append current MANIFEST and CURRENT file name
-  std::string file_name = DescriptorFileName("/titandb", manifest_file_number_);
-  files->push_back(file_name);
-  files->push_back(CurrentFileName("/titandb"));
+void BlobFileSet::GetAllFiles(std::vector<std::string>* files, 
+                              std::vector<VersionEdit>* edits) {
+  std::vector<std::string> all_blob_files;
 
-  // find current length of titandb manifest file while holding the mutex lock
-  *manifest_file_size = manifest_->file()->GetFileSize();
+  edits->clear();
+  edits->reserve(column_families_.size());
+  bool set_next_file_number = false;
+  for (auto& cf : column_families_) {
+    VersionEdit edit;
+    auto& blob_storage = cf.second;
+    // Get all blob files
+    blob_storage->GetAllFiles(&all_blob_files);
+    // Add all blob files to version_edit
+    edit.SetColumnFamilyID(cf.first);
+    if (!set_next_file_number) {
+      edit.SetNextFileNumber(next_file_number_.load());
+      set_next_file_number = true;
+    }
+    for (auto& file : blob_storage->files_) {
+      edit.AddBlobFile(file.second);
+    }
+    edits->emplace_back(edit);
+  }
+
+  files->clear();
+  files->reserve(all_blob_files.size() + 2);
+
+  for (auto& live_file : all_blob_files) {
+    files->emplace_back(live_file);
+  }
+
+  // Append current MANIFEST and CURRENT file name
+  files->emplace_back(DescriptorFileName("/titandb", manifest_file_number_));
+  files->emplace_back(CurrentFileName("/titandb"));
 }
 
 }  // namespace titandb
--- a/src/blob_file_set.cc
+++ b/src/blob_file_set.cc
@@ -335,18 +335,22 @@ void BlobFileSet::GetAllFiles(std::vector<std::string>* files,
 
   edits->clear();
   edits->reserve(column_families_.size());
-  bool set_next_file_number = false;
+
+  // Saves global information
+  {
+    VersionEdit edit;
+    edit.SetNextFileNumber(next_file_number_.load());
+    std::string record;
+    edit.EncodeTo(&record);
+    edits->emplace_back(edit);
+  }
+
+  // Saves all blob files
   for (auto& cf : column_families_) {
     VersionEdit edit;
-    auto& blob_storage = cf.second;
-    // Get all blob files
-    blob_storage->GetAllFiles(&all_blob_files);
-    // Add all blob files to version_edit
     edit.SetColumnFamilyID(cf.first);
-    if (!set_next_file_number) {
-      edit.SetNextFileNumber(next_file_number_.load());
-      set_next_file_number = true;
-    }
+    auto& blob_storage = cf.second;
+    blob_storage->GetAllFiles(&all_blob_files);
     for (auto& file : blob_storage->files_) {
       edit.AddBlobFile(file.second);
     }
--- a/src/blob_file_set.cc
+++ b/src/blob_file_set.cc
@@ -326,5 +326,26 @@ void BlobFileSet::GetObsoleteFiles(std::vector<std::string>* obsolete_files,
   obsolete_manifests_.clear();
 }
 
+void BlobFileSet::GetLiveFiles(std::vector<std::string>* live_files, uint64_t* manifest_file_size) {
+  for (auto it = column_families_.begin(); it != column_families_.end();) {
+    auto& cf_id = it->first;
+
+    if (obsolete_columns_.find(cf_id) == obsolete_columns_.end()) {
+      auto& blob_storage = it->second;
+      // get live blob file
+      blob_storage->GetLiveFiles(live_files);
+    }
+
+    ++it;
+  }
+  // append current MANIFEST and CURRENT file name
+  std::string file_name = DescriptorFileName("/titandb", next_file_number_.load()-1);
+  live_files->push_back(file_name);
+  live_files->push_back(CurrentFileName("/titandb"));
+
+  // find current length of titandb manifest file while holding the mutex lock
+  *manifest_file_size = manifest_->file()->GetFileSize();
+}
+
 }  // namespace titandb
 }  // namespace rocksdb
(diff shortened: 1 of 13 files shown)


caipengbo:
Thanks @yiwu-arbug for giving me advice on Slack

src/db_impl.cc
```diff
@@ -762,6 +763,80 @@ void TitanDBImpl::ReleaseSnapshot(const Snapshot* snapshot) {
...
+  if (!s.ok()) {
+    return s;
+  }
+
+  int count = 0;
+  {
+    // Hold delete_titandb_file_mutex_ to make sure no
+    // PurgeObsoleteFiles job is running.
+    MutexLock l(&delete_titandb_file_mutex_);
+    count = ++disable_titandb_file_deletions_;
```
Connor1996, line 778: can we use atomic value instead?

src/db_impl.cc
```diff
@@ -762,6 +763,80 @@ void TitanDBImpl::ReleaseSnapshot(const Snapshot* snapshot) {
...
+    return s;
+  }
+  
+  s = db_->GetLiveFiles(base_ret, base_manifest_file_size, flush_memtable);
+  if (!s.ok()) {
+    return s;
+  }
+
+  MutexLock l(&mutex_);
+  blob_file_set_->GetAllFiles(&titan_ret, titan_manifest_file_size);
```
Connor1996, line 835: After a second thought, Disbaling file deletion doesn't solve the problem. It's just keep the blob file from being pyhsically deleted, but blob files are logically deleted. Considing a Titan GC is happened between db_->GetLiveFiles and blob_file_set_->GetAllFiles, the titan_manifest_file_size and base_manifest_file_size are not fetched in atomic, so the blob file is logically deleted whereas rocksdb still refer to it. When reopen the checkpoint, titan will deleted the logical deleted file at start. Then it will meet a blob file not found issue.

src/titan_checkpoint_test.cc
```diff
@@ -0,0 +1,549 @@
...
+ public:
+  std::string dbname_;
+  std::string alternative_wal_dir_;
+  Env* env_;
+  TitanDB* db_;
+  TitanOptions last_options_;
+  std::vector<ColumnFamilyHandle*> handles_;
+  std::string snapshot_name_;
+
+  CheckpointTest() : env_(Env::Default()) {
```
Connor1996, line 33: please add a case that there is a Titan GC finished within the process of creating checkpoint.

src/db_impl.cc
caipengbo: NO, just like the comment said. If we use atomic value, there is no way to **prevent** `++disable_titandb_file_deletions_` while `TitanDBImpl::PurgeObsoleteFilesImpl` function is running.

src/titan_checkpoint_test.cc
caipengbo: Done

src/util.cc
```diff
@@ -166,5 +167,31 @@ Status SyncTitanManifest(Env* env, TitanStats* stats,
   return file->Sync(db_options->use_fsync);
 }
 
+Status CreateTitanManifest(Env* env, bool use_fsync,
```
Connor1996, line 170: Can we reuse it with `BlobFileSet::OpenManifest`?

caipengbo: It looks like `BlobFileSet::OpenManifest` has some overlap with `CreateTitanManifest`, but I'd like to have as little intrusion into Titan as possible

Connor1996: If that, let's move it out of util.cc

caipengbo: Ok, I moved it to titan_checkpoint_impl.cc

src/blob_file_set.cc
```diff
@@ -329,19 +329,36 @@ void BlobFileSet::GetObsoleteFiles(std::vector<std::string>* obsolete_files,
...
   }
-  // Append current MANIFEST and CURRENT file name
-  std::string file_name = DescriptorFileName("/titandb", manifest_file_number_);
-  files->push_back(file_name);
-  files->push_back(CurrentFileName("/titandb"));
 
-  // Find current length of titandb manifest file while holding the mutex lock
-  *manifest_file_size = manifest_->file()->GetFileSize();
+  // Append current MANIFEST and CURRENT file name
+  files->emplace_back(DescriptorFileName("/titandb", manifest_file_number_));
```
Connor1996, line 360: the origin manifest file should be excluded?

src/blob_file_set.cc
```diff
@@ -329,19 +329,36 @@ void BlobFileSet::GetObsoleteFiles(std::vector<std::string>* obsolete_files,
...
+  edits->reserve(column_families_.size());
+  
+  for (auto& cf : column_families_) {
+    VersionEdit edit;
+    auto& blob_storage = cf.second;
+    // Get all blob files
+    blob_storage->GetAllFiles(&all_blob_files);
+    // Add all blob files to version_edit
+    edit.SetColumnFamilyID(cf.first);
+    edit.SetNextFileNumber(next_file_number_.load());
```
Connor1996, line 345: no need to set next file number everytime

src/blob_file_set.cc
caipengbo: Oh, that's true

caipengbo: No, we only obtain the name of the file, we use the old `manifest_file_number_`, but the content is produced by ourselves. If we excluded it, we will lose the manifest file name.

src/blob_file_set.cc
caipengbo: Ok,  I will set next file number in first time.

src/titan_checkpoint_impl.cc
```diff
@@ -0,0 +1,392 @@
...
+  std::vector<std::string> titan_live_files;
+  std::vector<VersionEdit> version_edits;
+  uint64_t base_manifest_file_size = 0;
+  uint64_t min_log_num = port::kMaxUint64;
+  *sequence_number = db_->GetLatestSequenceNumber();
+  bool same_fs = true;
+  VectorLogPtr live_wal_files;
+
+  bool flush_memtable = true;
+  if (!db_options.allow_2pc) {
```
Connor1996, line 207: Can we make the process like that:
1. DisableFileDeletions, this will prevent blob files being deleted
2. Call db_->CreateCustomCheckpoint without mutex to get a checkpoint of base db
3. Collect all file（live + obsolete）about titandb and record all blob files into version_edit(prevent removing useful blob file when Titan reopen)
4. EnableFileDeletions

If that, we can reuse the logic of `CreateCustomCheckpoint` of RocksDB

src/titan_checkpoint_impl.cc
```diff
@@ -306,9 +305,13 @@ Status TitanCheckpointImpl::CreateCustomCheckpoint(
     }
     if ((type != kTableFile && type != kBlobFile) || (!same_fs)) {
       if (type == kDescriptorFile) {
-        s = copy_file_cb(db_->GetName(), src_fname,
-                (in_titan_dir) ? titan_manifest_file_size : base_manifest_file_size,
-                type);  
+        if (in_titan_dir) {
+          // Craft titan manifest file, ensure include all titan file.
+          CreateTitanManifest(db_->GetEnv(), db_->GetDBOptions().use_fsync,
```
Connor1996: it doesn't check the returned status

src/titan_checkpoint_impl.cc
caipengbo: I reused the the logic of `CreateCustomCheckpoint` in new commit, and modified `CurrentFileModifiedWhileCheckpointing` test case.

src/titan_checkpoint_impl.cc
caipengbo: Fixed it.

src/titan_checkpoint_impl.cc
```diff
@@ -194,195 +195,78 @@ Status TitanCheckpointImpl::CreateCustomCheckpoint(
...
-      live_wal_files.clear();
-    }
-  }
-
-  // This will return live files prefixed with "/"
-  s = db_->GetTitanLiveFiles(live_files, &base_manifest_file_size,
-                              &version_edits, flush_memtable);
+  // Create base db checkpoint
+  auto base_db_checkpoint = new rocksdb::CheckpointImpl(db_);
+  s = base_db_checkpoint->CreateCustomCheckpoint(db_options, link_file_cb,
```
Connor1996, line 204: it should be called between `DisableFileDeletions` and `EnableFileDeletions`

caipengbo: On the outside of `TitanCheckpointImpl::CreateCustomCheckpoint`, we've already called `DisableFileDeletions ` and `EnableFileDeletions `. I think we can skip it here.

Connor1996: Then is it necessary to call `DisableFileDeletions` and `EnableFileDeletions` again in `GetAllTitanFiles`?

caipengbo: `GetAllTitanFiles` may be called by other functions in the future, but I think `CreateCustomCheckPoint` will only be used by `CreateCheckPoint`.

Connor1996: Would someone will call `CreateCustomCheckpoint` directly? If that, the behavior would be broken.

caipengbo: Done.

Connor1996: Seems the `EnableFileDeletions` shoud be called after `GetAllTitanFiles`?

Connor1996: And we don't need to call `DisableFileDeletions` and `EnableFileDeletions` in `CreateCheckpoint` anymore.

caipengbo: Oh, that's true, I fixed it.

src/titan_checkpoint_impl.cc
```diff
@@ -0,0 +1,292 @@
...
+  CleanStagingDirectory(full_private_path, db_options.info_log.get());
+  // Create checkpoint directory and subdirectory
+  s = db_->GetEnv()->CreateDir(full_private_path);
+  if (s.ok()) {
+    s = db_->GetEnv()->CreateDir(full_private_path + "/titandb");
+  }
+  uint64_t sequence_number = 0;
+  if (s.ok()) {
+    // Disable file deletions
+    s = db_->DisableFileDeletions();
```
Connor1996, line 123: don't need it anymore

Connor1996 (commented):
Please fix the sign-off and format

src/titan_checkpoint_impl.cc
caipengbo: If we remove `DisableFileDeletions` and `EnableFileDeletions` in CreateCheckpoint , we should move `EnableFileDeletions ` to the last in `CreateCustomCheckpoint`, which will prevent delete files when link or copy files. If we do it in this way, we're essentially moving `DisableFileDeletions` and `EnableFileDeletions` into `CreateCheckpoint `.

Connor1996:
The CI is still failing, please check it

src/titan_checkpoint_impl.cc
```diff
@@ -0,0 +1,300 @@
...
+      assert(ss.ok());
+    }
+    return s;
+  }
+
+  // This will return files prefixed with "/titandb"
+  s = db_->GetAllTitanFiles(titandb_files, &version_edits);
+
+  if (disabled_file_deletions) {
+    Status ss = db_->EnableFileDeletions(false);
```
Connor1996, line 227: This should be called after linking or copying the files. And I check the RocksDB's code, seems it just calls `EnableFileDeletions` and `DisableFileDeletions` in `CreateCheckpoint`. So let's just remove these(sorry about that)

Connor1996 (commented):
rest LGTM

Connor1996:
PTAL @yiwu-arbug

src/titan_checkpoint_impl.cc
caipengbo: Ok, I'll change it right away.

src/blob_file_set.cc
```diff
@@ -326,5 +329,41 @@ void BlobFileSet::GetObsoleteFiles(std::vector<std::string>* obsolete_files,
...
+  bool set_next_file_number = false;
+  for (auto& cf : column_families_) {
+    VersionEdit edit;
+    auto& blob_storage = cf.second;
+    // Get all blob files
+    blob_storage->GetAllFiles(&all_blob_files);
+    // Add all blob files to version_edit
+    edit.SetColumnFamilyID(cf.first);
+    if (!set_next_file_number) {
+      edit.SetNextFileNumber(next_file_number_.load());
```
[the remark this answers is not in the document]
caipengbo: Done

Connor1996 (approved):
LGTM

src/blob_file_set.cc
```diff
@@ -326,5 +329,45 @@ void BlobFileSet::GetObsoleteFiles(std::vector<std::string>* obsolete_files,
...
+
+  files->clear();
+  files->reserve(all_blob_files.size() + 2);
+
+  for (auto& live_file : all_blob_files) {
+    files->emplace_back(live_file);
+  }
+
+  // Append current MANIFEST and CURRENT file name
+  files->emplace_back(DescriptorFileName("/titandb", manifest_file_number_));
```
yiwu-arbug, line 368: should the path be consistent with `TitanDBOptions::dirname`?

src/blob_storage.cc
```diff
@@ -187,6 +187,16 @@ void BlobStorage::GetObsoleteFiles(std::vector<std::string>* obsolete_files,
   }
 }
 
+void BlobStorage::GetAllFiles(std::vector<std::string>* files) {
+  MutexLock l(&mutex_);
+
+  for (auto& file : files_) {
+    uint64_t file_number = file.first;
+    // relative to dbname, the form like: "/titandb/[0-9].blob"
+    files->emplace_back(BlobFileName("", "titandb", file_number));
```
yiwu-arbug, line 196: should the path be consistent with `TitanDBOptions::dirname`?

yiwu-arbug (commented):
Looking good overall. Minor comments.

src/blob_file_set.cc
caipengbo: I found my error, my current implementation only works when `TitanDBOptions::dirname` is the default value, when it is not the default value, will generate an error. I will fix the problem.

yiwu-arbug (approved):
LGTM

ti-srebot (approved):

yiwu-arbug:
Defer to @Connor1996 for detail review of latest changes.

src/titan_checkpoint_impl.cc
```diff
@@ -83,20 +85,29 @@ Status TitanCheckpointImpl::CreateTitanManifest(
 }
 
 // Builds an openable checkpoint of TitanDB
-Status TitanCheckpointImpl::CreateCheckpoint(const std::string& checkpoint_dir,
-                                             uint64_t log_size_for_flush) {
-  DBOptions db_options = db_->GetDBOptions();
+Status TitanCheckpointImpl::CreateCheckpoint(
+              const std::string& base_checkpoint_dir,
+              const std::string& titan_checkpoint_dir,
```
Connor1996: Can we get the titan relative dir by the `dirname_` in `TitanDBImpl`? So we can construct the titan checkpoint dir instead of specifying one.

caipengbo: I've thought about this before, but I think `dirname_` can be specified by the user to any absolute path, it's only specified as a relative location by default.

src/titan_checkpoint_impl.cc
```diff
@@ -0,0 +1,288 @@
...
+// Builds an openable checkpoint of TitanDB
+Status TitanCheckpointImpl::CreateCheckpoint(
+    const std::string& base_checkpoint_dir,
+    const std::string& titan_checkpoint_dir, uint64_t log_size_for_flush) {
+  TitanDBOptions titandb_options = db_->GetTitanDBOptions();
+  std::string full_private_path;
+  std::string checkpoint_dir = base_checkpoint_dir + "/titandb";
+  if (!titan_checkpoint_dir.empty()) {
+    checkpoint_dir = titan_checkpoint_dir;
+  }
```
Connor1996, line 96:
```suggestion
  std::string checkpoint_dir = titan_checkpoint_dir;
  if (!checkpoint_dir.empty()) {
    checkpoint_dir =  base_checkpoint_dir + "/titandb";
  }
```

Connor1996 (commented):
rest LGTM

src/titan_checkpoint_impl.cc
caipengbo: Done!
PS: shoud `if (checkpoint_dir.empty()) {`

PR #210, opened 2021-04-23
Fix the titan invalid ratelimiter in Flush/GC

### LINKED ISSUE #209: Ratelimiter is invalid in write blobfile with Flush and GC
There is a bug that i have find in write blobfile when we want to limit the write blobfile 's rate during Flush and GC.

Test titandb_bench
```shell
./titandb_bench \
  --benchmarks="fillrandom,stats" \
  --max_background_compactions=32 \
  --max_background_flushes=4 \
  --max_write_buffer_number=6 \
  --target_file_size_base=67108864 \
  --max_bytes_for_level_base=536870912 \
  --statistics=true \
  --stats_dump_period_sec=5 \
  --num=5000000 \
  --duration=300 \
  --threads=10 \
  --value_size=8192 \
  --key_size=16 \
  --enable_pipelined_write=false \
  --db=./db_bench_test \
  --wal_dir=./db_bench_test \
  --num_multi_db=1 \
  --allow_concurrent_memtable_write=true \
  --disable_wal=true \ # Let the write db only have the flush and compaction I/O
  --use_titan=true \
  --titan_max_background_gc=2 \
  --rate_limiter_bytes_per_sec=134217728 \
  --rate_limiter_auto_tuned=false 
```
1. First test is `use_titan=false` with `rate_limiter_bytes_per_sec=134217728`，the disk write rate is limited:

  image
  
2. Second test is `use_titan=true` with `rate_limiter_bytes_per_sec=134217728`, but the disk write rate is not limited:
 
image

The reason why titan is not limited in write blobfile with Flush and GC is that `io_priority` in `RequestToken` is not satisfy.
Titan use the default WritableFile's io_priority_ which is `Env::IO_TOTAL`。
```c
size_t RateLimiter::RequestToken(size_t bytes, size_t alignment,
                                 Env::IOPriority io_priority, Statistics* stats,
                                 RateLimiter::OpType op_type) {
  if (io_priority < Env::IO_TOTAL && IsRateLimited(op_type)) {
  ...
  }
```

BaronStack:
Fix #209 ;

And the different `io_priority_` when  invoke a`NewFile` to create  the blobfile.

We always want limit the compaction/gc's write IO to keep the latency of read, so i keep the Flush with `IO_HIGH` pri and set the GC wich `IO_LOW` pri.

zhanghuigui@kuaishou.com

### DIFF
--- a/src/blob_file_manager.h
+++ b/src/blob_file_manager.h
@@ -26,7 +26,8 @@ class BlobFileManager {
   // Creates a new file. The new file should not be accessed until
   // FinishFile() has been called.
   // If successful, sets "*handle* to the new file handle.
-  virtual Status NewFile(std::unique_ptr<BlobFileHandle>* handle) = 0;
+  virtual Status NewFile(std::unique_ptr<BlobFileHandle>* handle,
+                         Env::IOPriority pri = Env::IOPriority::IO_LOW) = 0;
 
   // Finishes the file with the provided metadata. Stops writting to
   // the file anymore.
--- a/src/blob_file_manager.h
+++ b/src/blob_file_manager.h
@@ -26,6 +26,7 @@ class BlobFileManager {
   // Creates a new file. The new file should not be accessed until
   // FinishFile() has been called.
   // If successful, sets "*handle* to the new file handle.
+  // And also, sets the IOPriority to the new file handle.
   virtual Status NewFile(std::unique_ptr<BlobFileHandle>* handle,
                          Env::IOPriority pri = Env::IOPriority::IO_LOW) = 0;
 
--- a/src/blob_file_manager.h
+++ b/src/blob_file_manager.h
@@ -25,10 +25,14 @@ class BlobFileManager {
 
   // Creates a new file. The new file should not be accessed until
   // FinishFile() has been called.
-  // If successful, sets "*handle* to the new file handle.
-  // And also, sets the IOPriority to the new file handle.
+  // If successful, sets "*handle* to the new file handle with given
+  // IOPriority.
+  //
+  // The reason why set the io priority for WritableFile in Flush,
+  // Compaction and GC is that the ratelimiter will use the default
+  // priority IO_TOTAL which won't be limited in ratelimiter.
   virtual Status NewFile(std::unique_ptr<BlobFileHandle>* handle,
-                         Env::IOPriority pri = Env::IOPriority::IO_LOW) = 0;
+                         Env::IOPriority pri = Env::IOPriority::IO_TOTAL) = 0;
 
   // Finishes the file with the provided metadata. Stops writting to
   // the file anymore.
--- a/src/blob_gc_job.cc
+++ b/src/blob_gc_job.cc
@@ -211,7 +211,7 @@ Status BlobGCJob::DoRunGC() {
             std::move(blob_file_handle), std::move(blob_file_builder)));
       }
       // Set the GC's blob file with a low_io pri in ratelimiter
-      s = blob_file_manager_->NewFile(&blob_file_handle, 
+      s = blob_file_manager_->NewFile(&blob_file_handle,
                                       Env::IOPriority::IO_LOW);
       if (!s.ok()) {
         break;
--- a/src/blob_gc_job.cc
+++ b/src/blob_gc_job.cc
@@ -210,7 +210,9 @@ Status BlobGCJob::DoRunGC() {
         blob_file_builders_.emplace_back(std::make_pair(
             std::move(blob_file_handle), std::move(blob_file_builder)));
       }
-      s = blob_file_manager_->NewFile(&blob_file_handle);
+      // Set the GC's blob file with a low_io pri in ratelimiter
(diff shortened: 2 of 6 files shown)


BaronStack:
The reason why the test is not successful is that i modify the ratelimiter unitest, original is not work in check the real ratelimiter.

src/table_builder.cc
```diff
@@ -146,7 +146,8 @@ void TitanTableBuilder::AddBlob(const ParsedInternalKey& ikey,
 
   // Init blob_builder_ first
   if (!blob_builder_) {
-    status_ = blob_manager_->NewFile(&blob_handle_);
+    // Set the GC's blob file with a high_io pri in ratelimiter
```
Connor1996, line 149: Seems the comment is wrong. It should be `Flush's`?

BaronStack: yes ,you're right.

src/table_builder.cc
```diff
@@ -146,7 +146,8 @@ void TitanTableBuilder::AddBlob(const ParsedInternalKey& ikey,
 
   // Init blob_builder_ first
   if (!blob_builder_) {
-    status_ = blob_manager_->NewFile(&blob_handle_);
+    // Set the Flush's blob file with a high_io pri in ratelimiter
+    status_ = blob_manager_->NewFile(&blob_handle_, Env::IOPriority::IO_HIGH);
```
Connor1996, line 150: the blob file may be generated during compaction, for compaction seems we should set low pri? you can check `target_level_` to decide if it's a flush or compaction

Connor1996:
Please fix the sign-off. You should resign all the previous commits.

Connor1996:
Also, welcome to join #sig-engine on slack https://tikv-wg.slack.com/archives/CQ545HDGC to discuss

src/table_builder_test.cc
```diff
@@ -24,14 +24,16 @@ class FileManager : public BlobFileManager {
         number_(kTestFileNumber),
         blob_file_set_(blob_file_set) {}
 
-  Status NewFile(std::unique_ptr<BlobFileHandle>* handle) override {
+  Status NewFile(std::unique_ptr<BlobFileHandle>* handle,
+                 Env::IOPriority pri = Env::IOPriority::IO_LOW) override {
```
tabokie, line 28:
```suggestion
                 Env::IOPriority pri = Env::IOPriority::IO_TOTAL) override {
```

src/blob_file_manager.h
```diff
@@ -26,7 +26,9 @@ class BlobFileManager {
   // Creates a new file. The new file should not be accessed until
   // FinishFile() has been called.
   // If successful, sets "*handle* to the new file handle.
```
tabokie, line 28:
```suggestion
  // If successful, sets "*handle* to the new file handle with given
  // IOPriority.
```

src/blob_file_manager.h
```diff
@@ -26,7 +26,9 @@ class BlobFileManager {
...
-  virtual Status NewFile(std::unique_ptr<BlobFileHandle>* handle) = 0;
+  // And also, sets the IOPriority to the new file handle.
+  virtual Status NewFile(std::unique_ptr<BlobFileHandle>* handle,
+                         Env::IOPriority pri = Env::IOPriority::IO_LOW) = 0;
```
tabokie, line 31:
```suggestion
                         Env::IOPriority pri = Env::IOPriority::IO_TOTAL) = 0;
```

tabokie: It's consistent with RocksDB to keep default value as noop.

src/blob_gc_job.cc
```diff
@@ -210,7 +210,9 @@ Status BlobGCJob::DoRunGC() {
         blob_file_builders_.emplace_back(std::make_pair(
             std::move(blob_file_handle), std::move(blob_file_builder)));
       }
-      s = blob_file_manager_->NewFile(&blob_file_handle);
+      // Set the GC's blob file with a low_io pri in ratelimiter
```
tabokie, line 213: probably don't needed this comment, the code is self-explanatory.

src/db_impl.cc
```diff
@@ -39,6 +40,12 @@ class TitanDBImpl::FileManager : public BlobFileManager {
       std::unique_ptr<WritableFile> f;
       s = db_->env_->NewWritableFile(name, &f, db_->env_options_);
       if (!s.ok()) return s;
+
+      // Set the io priority for WritableFile in Flush or GC.
+      // Or the ratelimiter will use the default priority IO_ALL
+      // which won't be limited in ratelimiter.
```
tabokie, line 46: Instead of here, I think it's better to explain the necessity to pass in IOPriority at `NewFile` declaration in src/blob_file_manager.h

tabokie (commented):
Minor ones

BaronStack:
@tabokie Thanks for your advise!

tabokie (approved):
Looking good

Connor1996 (approved):
LGTM

ti-srebot (approved):

yiwu-arbug (approved):

PR #212, opened 2021-05-19
Fix missing obsolete sequence for deletion in deleteblobfiles

Connor1996:
When doing DeleteBlobFilesInRange, it shouldn't purge the blob files immediately, otherwise, the existing snapshots will be broken and cause `Missing Blob` errors.

image

To solve that, Titan introduces the obsolete sequence to make sure the blob files are purged only after the snapshots are released.  But in `DeleteBlobFilesInRange`, the obsolete sequence is not set, which breaks the promise.

### DIFF
--- a/src/blob_file_set.cc
+++ b/src/blob_file_set.cc
@@ -287,7 +287,7 @@ Status BlobFileSet::DeleteBlobFilesInRanges(uint32_t cf_id,
     Status s = it->second->GetBlobFilesInRanges(ranges, n, include_end, &files);
     if (!s.ok()) return s;
     for (auto file_number : files) {
-      edit.DeleteBlobFile(file_number);
+      edit.DeleteBlobFile(file_number, obsolete_sequence);
     }
     s = LogAndApply(edit);
     return s;
--- a/src/titan_db_test.cc
+++ b/src/titan_db_test.cc
@@ -728,6 +728,10 @@ TEST_F(TitanDBTest, DeleteFilesInRange) {
   ASSERT_TRUE(db_->GetProperty("rocksdb.num-files-at-level6", &value));
   ASSERT_EQ(value, "3");
 
+  std::unique_ptr<Iterator> iter(db_->NewIterator(ReadOptions()));
+  iter->SeekToFirst();
+  ASSERT_TRUE(iter->Valid());
+
   std::string key40 = GenKey(40);
   std::string key80 = GenKey(80);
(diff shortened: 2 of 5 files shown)


tabokie (approved):

yiwu-arbug (approved):

ti-srebot (approved):

gotoxu:
@Connor1996  我们的集群也出现了这个问题，将TiKV升级到4.0.14以后仍然会不停Panic(毕竟文件已经被删除了)，请问一下这种异常应该如何恢复？有没有什么好的办法？

Connor1996:
@gotoxu What's the version of the previous Titan? It shouldn't be this bug, this issue will be fixed after a restart.

gotoxu:
The previous version was TiKV 4.0.12.

Before upgrading to 4.0.14, TiKV has been Panic, and the exception information and call stack are the same as the information in the image you uploaded.

After updating to 4.0.14, we restarted TiKV, but it still keeps panic. So we removed this node and re-expanded a new one.

Now, the new node will still report Missing blob file error, but it is no longer panic, the error message is as follows:
```shell
[2021/08/08 01:10:10.411 +08:00] [ERROR] [region_snapshot.rs:236] ["failed to get value of key in cf"] [error="Engine(\"Corruption: Missing blob file: 840455\")"] [cf=default] [region=122896] [key=67616D652D757365FF722D636F6E74656EFF742D707573685F32FF373438343A303030FF3A4D3A67616D655FFF636F6E74656E745FFF707573685F76303AFF3836393934303033FF3738363631313500FEFA13ABB2A747FE03]
[2021/08/08 01:10:10.411 +08:00] [ERROR] [errors.rs:316] ["txn aborts"] [err_code=KV:Unknown] [err="Txn(Mvcc(Engine(Other(\"[src/server/raftkv.rs:419]: Storage Engine Corruption: Missing blob file: 840455\"))))"]
```

Connor1996:
@gotoxu though the call stack is the same, the root reason maybe not same. From the place where the error occurs, it seems not the bug that this PR solved.

Connor1996:
@gotoxu Is the env still there, may I take a look?

gotoxu:
The log of TiKV 4.0.12 has been cleaned up.

Here is the log after upgrading to TiKV 4.0.14: 

[tikv.err.log.zip](https://github.com/tikv/titan/files/7007163/tikv.err.log.zip)

Hope it can help you!

PR #221, opened 2021-08-19
Fix false corruption error when reading from v1 blob file

yiwu-arbug:
To support dictionary compression we added a v2 blob file header to store extra flags. However, the check in BlobFileReader , when reading a v1 blob file, falsely asserted the file header is shorter than expected:
https://github.com/tikv/titan/blob/ecb5cba016096309cbd9b566bdc4ce62307527a0/src/blob_file_reader.cc#L103
https://github.com/tikv/titan/blob/ecb5cba016096309cbd9b566bdc4ce62307527a0/src/blob_format.h#L394
Refactoring the code to bypass the check to fix the issue.

The issue is introduced in https://github.com/tikv/titan/pull/189. TiKV is affected when Titan is enabled and upgrade from pre-5.0 versions to >=5.0.0 versions. It will make TiKV fall in crash loop.

Also adding a titan_blob_file_dump tool to dump blob file content.

### DIFF
--- a/src/blob_file_iterator.cc
+++ b/src/blob_file_iterator.cc
@@ -28,7 +28,7 @@ bool BlobFileIterator::Init() {
     return false;
   }
   BlobFileHeader blob_file_header;
-  status_ = DecodeInto(slice, &blob_file_header, true/*ignore_extra_bytes*/);
+  status_ = DecodeInto(slice, &blob_file_header, true /*ignore_extra_bytes*/);
   if (!status_.ok()) {
     return false;
   }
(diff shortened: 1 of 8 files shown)


src/blob_file_iterator.cc
```diff
@@ -28,7 +28,7 @@ bool BlobFileIterator::Init() {
     return false;
   }
   BlobFileHeader blob_file_header;
-  status_ = blob_file_header.DecodeFrom(&slice);
+  status_ = DecodeInto(slice, &blob_file_header, true/*ignore_extra_bytes*/);
```
tabokie, line 31:
```suggestion
  status_ = DecodeInto(slice, &blob_file_header, true /*ignore_extra_bytes*/);
```

tabokie (approved):
LGTM

PR #223, opened 2021-08-31
Fix false corruption error when reading from v1 blob file (#221)

yiwu-arbug:
cherry-picking #221 

To support dictionary compression we added a v2 blob file header to store extra flags. However, the check in BlobFileReader , when reading a v1 blob file, falsely asserted the file header is shorter than expected:
https://github.com/tikv/titan/blob/ecb5cba016096309cbd9b566bdc4ce62307527a0/src/blob_file_reader.cc#L103
https://github.com/tikv/titan/blob/ecb5cba016096309cbd9b566bdc4ce62307527a0/src/blob_format.h#L394
Refactoring the code to bypass the check to fix the issue.

The issue is introduced in https://github.com/tikv/titan/pull/189. TiKV is affected when Titan is enabled and upgrade from pre-5.0 versions to >=5.0.0 versions. It will make TiKV fall in crash loop.

Also adding a titan_blob_file_dump tool to dump blob file content.

### DIFF
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -5,26 +5,38 @@ namespace titandb {
 
 BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
                                  const TitanCFOptions& cf_options,
-                                 WritableFileWriter* file)
+                                 WritableFileWriter* file,
+                                 uint32_t blob_file_version)
     : builder_state_(cf_options.blob_file_compression_options.max_dict_bytes > 0
                          ? BuilderState::kBuffered
                          : BuilderState::kUnbuffered),
       cf_options_(cf_options),
       file_(file),
+      blob_file_version_(blob_file_version),
       encoder_(cf_options.blob_file_compression,
                cf_options.blob_file_compression_options) {
-#if ZSTD_VERSION_NUMBER < 10103
-  if (cf_options_.blob_file_compression_options.max_dict_bytes > 0) {
-    status_ = Status::NotSupported("ZSTD version too old.");
+  status_ = BlobFileHeader::ValidateVersion(blob_file_version_);
+  if (!status_.ok()) {
     return;
   }
+  if (cf_options_.blob_file_compression_options.max_dict_bytes > 0) {
+    if (blob_file_version_ != BlobFileHeader::kVersion2) {
+      status_ = Status::NotSupported(
+          "dictionary comparession is not supported by blob file version 1");
+    }
+#if ZSTD_VERSION_NUMBER < 10103
+    status_ = Status::NotSupported("ZSTD version too old.");
+    return;
 #endif
+  }
   WriteHeader();
 }
 
 void BlobFileBuilder::WriteHeader() {
   BlobFileHeader header;
+  header.version = blob_file_version_;
   if (cf_options_.blob_file_compression_options.max_dict_bytes > 0) {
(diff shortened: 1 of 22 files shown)


src/titan_db_test.cc
```diff
@@ -1987,7 +1987,7 @@ TEST_F(TitanDBTest, Config) {
 }
 
 #if defined(__linux) && !defined(TRAVIS)
-TEST_F(TitanDBTest, NoSpaceLeft) {
+TEST_F(TitanDBTest, DISABLED_NoSpaceLeft) {
```
Connor1996, line 1990: why disable it

Connor1996 (approved):
LGTM

src/titan_db_test.cc
yiwu-arbug: keep failing in CI environment. we disabled it previously on travis, too.

yiwu-arbug: though it would be great if we can fix it.

src/blob_file_builder.cc
```diff
@@ -5,26 +5,38 @@ namespace titandb {
...
                cf_options.blob_file_compression_options) {
-#if ZSTD_VERSION_NUMBER < 10103
+  status_ = BlobFileHeader::ValidateVersion(blob_file_version_);
+  if (!status_.ok()) {
+    return;
+  }
   if (cf_options_.blob_file_compression_options.max_dict_bytes > 0) {
+    if (blob_file_version_ != BlobFileHeader::kVersion2) {
+      status_ = Status::NotSupported(
+          "dictionary comparession is not supported by blob file version 1");
```
tabokie, line 25:
```suggestion
          "dictionary compression is not supported by blob file version 1");
```

src/titan_logging.h
```diff
@@ -0,0 +1,71 @@
...
+                       TitanLogShorterFileName(__FILE__), ##__VA_ARGS__)
+
+#define TITAN_LOG_BUFFER_MAX_SZ(LOG_BUF, MAX_LOG_SIZE, FMT, ...) \
+  rocksdb::LogToBuffer(LOG_BUF, MAX_LOG_SIZE,                    \
+                       TITAN_LOG_PREPEND_FILE_LINE(FMT),         \
+                       TitanLogShorterFileName(__FILE__), ##__VA_ARGS__)
+
+#define TITAN_LOG_DETAILS(LGR, FMT, ...) \
+  ;  // due to overhead by default skip such lines
+// TITAN_LOG_DEBUG(LGR, FMT, ##__VA_ARGS__)
```
tabokie, line 71: redundant?

tabokie (approved):
LGTM

ti-srebot (approved):

src/titan_logging.h
yiwu-arbug: need this to pass CI.

yiwu-arbug: just cherry-picking the change from master. I will update master instead.

src/blob_file_builder.cc
yiwu-arbug: It is a cherry-pick. I will update master instead.

src/titan_logging.h
yiwu-arbug: This is a copy-paste from rocksdb/logging/logging.h. I will keep it instead.

PR #233, opened 2022-03-22
Fix a problem in TitanTableBuilder::Add()

### LINKED ISSUE #232: An issues about TitanTableBuilder::Add()
I am reading the source code of titan, and I found an issue.

In TitanTableBuilder::Add() (pasted in below), there are several "if" & "else" statements, and I want to talking about the assertion in that.
In the last "else", there is an assertion that expecting builder_unbuffered() returns true.
But we can assume such a scenario: 
1. There is a memtable to be flushed, and it includes several **big kvs (i.e., has big value)** and some **blob indexes (i.e., the type of their key is kTypeBlobIndex)** generated by GC. 
2. When flushing, a few big kvs are added to the TitanTableBuilder first with TitanTableBuilder::Add (the below function), which add them in a blob file, and because the number of them is few, the state of BlobfileBuilder is still Buffered.  
3. Then, add a blob index with TitanTableBuilder::Add(), and it will reach the last "else". In this case, because the state of BlobfileBuilder is Buffered, the assertion will fail.

I'm not sure whether my understanding is correct, and it looks like a potential problem.

```
/* titan/src/table_builder.cc */

void TitanTableBuilder::Add(const Slice& key, const Slice& value){

  if (ikey.type == kTypeBlobIndex &&
      cf_options_.blob_run_mode == TitanBlobRunMode::kFallback) {

      ....// processing something only when fallback

   }else if (ikey.type == kTypeValue &&
             cf_options_.blob_run_mode == TitanBlobRunMode::kNormal){

      ....// processing something when encountering kTypeValue keys

   }else if (ikey.type == kTypeBlobIndex && cf_options_.level_merge &&
             target_level_ >= merge_level_ &&
             cf_options_.blob_run_mode == TitanBlobRunMode::kNormal) {

      ....// processing something when running level merge in compaction

   } else { 
    // processing when encountering kTypeBlobIndex or kTypeMerge
    assert(builder_unbuffered()); 
    base_builder_->Add(key, value);
  }
}
```

Borelset:
Close #232.

Fix a problem in a corner case:
Flushing a memtable, and it has a few kTypeValue in the first and follows with a kTypeBlobIndex or kTypeMerge generated by GC.
These kTypeValues are written to a blob file first, which makes the BlobFileBuilder is still in the Buffered state because the size of them is not enough.
In this case, the following kTypeBlobIndex or kTypeMerge must be cached in the workflow of TitanTableBuilder::Add().

### DIFF
--- a/src/table_builder.cc
+++ b/src/table_builder.cc
@@ -70,7 +70,7 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
              cf_options_.blob_run_mode == TitanBlobRunMode::kNormal) {
     bool is_small_kv = value.size() < cf_options_.min_blob_size;
     if (is_small_kv) {
-      AddSmall(key, ikey, value);
+      AddBase(key, ikey, value);
       return;
     } else {
       // We write to blob file and insert index
@@ -106,17 +106,17 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
                         index.file_number, get_status.ToString().c_str());
       }
     }
-    AddSmall(key, ikey, value);
+    AddBase(key, ikey, value);
   } else {
     // Mainly processing kTypeMerge and kTypeBlobIndex in both flushing and
     // compaction.
-    AddSmall(key, ikey, value);
+    AddBase(key, ikey, value);
   }
 }
 
-void TitanTableBuilder::AddSmall(const Slice& key,
-                                 const ParsedInternalKey& parsedKey,
-                                 const Slice& value) {
+void TitanTableBuilder::AddBase(const Slice& key,
+                                const ParsedInternalKey& parsedKey,
+                                const Slice& value) {
   if (builder_unbuffered()) {
     // We can directly append this into SST safely, without disorder issue.
     // Only when base_builder_ is in unbuffered state
@@ -186,10 +186,10 @@ void TitanTableBuilder::AddBlob(const ParsedInternalKey& ikey,
     FinishBlobFile();
   }
 
-  AddToBaseTable(contexts);
+  AddBlobResultsToBase(contexts);
 }
 
-void TitanTableBuilder::AddToBaseTable(
+void TitanTableBuilder::AddBlobResultsToBase(
     const BlobFileBuilder::OutContexts& contexts) {
   if (contexts.empty()) return;
   for (const std::unique_ptr<BlobFileBuilder::BlobRecordContext>& ctx :
@@ -229,7 +229,7 @@ void TitanTableBuilder::FinishBlobFile() {
     s = blob_builder_->Finish(&contexts);
     UpdateIOBytes(prev_bytes_read, prev_bytes_written, &io_bytes_read_,
                   &io_bytes_written_);
-    AddToBaseTable(contexts);
+    AddBlobResultsToBase(contexts);
 
     if (s.ok() && ok()) {
       TITAN_LOG_INFO(db_options_.info_log,
--- a/src/table_builder.cc
+++ b/src/table_builder.cc
@@ -183,7 +183,8 @@ void TitanTableBuilder::AddBlob(const ParsedInternalKey& ikey,
       cf_options_.blob_file_target_size) {
     // if blob file hit the size limit, we have to finish it
     // in this case, when calling `BlobFileBuilder::Finish`, builder will be in
-    // unbuffered state, so it will not trigger another `AddBlobResultsToBase` call
+    // unbuffered state, so it will not trigger another `AddBlobResultsToBase`
+    // call
     FinishBlobFile();
   }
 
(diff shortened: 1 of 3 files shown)


Connor1996 (commented):
Please add a test case

Borelset:
> Please add a test case

Ok, I have updated the "DictCompressDisorder" test in titan/src/table_builder_test.cc to cover this case.

Connor1996 (approved):
LGTM, Please fix the format check and sign-off

src/table_builder.cc
```diff
@@ -122,9 +122,16 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+  } else {// Mainly processing kTypeMerge and kTypeBlobIndex in both flushing and compaction.
+      if (builder_unbuffered()) {
+          // Directly append when base_builder_ is in unbuffered state
+          base_builder_->Add(key, value);
+      } else {
+          //It must be cached when base_builder_ is in buffered state
+          std::unique_ptr<BlobFileBuilder::BlobRecordContext> ctx =
+                  NewCachedRecordContext(ikey, value);
+          blob_builder_->AddSmall(std::move(ctx));
+      }
```
Connor1996, line 134: it's repeated for third times, how about extracting a function.

src/table_builder.cc
```diff
@@ -115,16 +106,24 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+    AddSmallToTableAdaptively(ikey, value);
   } else {
-    assert(builder_unbuffered());
-    base_builder_->Add(key, value);
+    // Mainly processing kTypeMerge and kTypeBlobIndex in both flushing and compaction.
+    AddSmallToTableAdaptively(ikey, value);
+  }
+}
+
+void TitanTableBuilder::AddSmallToTableAdaptively(const ParsedInternalKey& ikey, const Slice& value){
```
Connor1996, line 116: Rename it to `AddToBaseTable`, and change L59 and L67 too

src/table_builder.cc
```diff
@@ -115,16 +106,24 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+  if (builder_unbuffered()) {
+    // We can directly append this into SST safely, without disorder issue.
+    // Only when base_builder_ is in unbuffered state
+    base_builder_->Add(ikey.user_key, value);
+  } else {
+    // We have to let builder to cache this KV pair, and it will be returned
+    // when state changed
```
Connor1996:
```suggestion
    // We have to let builder to cache this KV pair, and it will be flushed to base table
    // when the state changes to unbuffered
```

Borelset:
There seems to be some problem with the sign-off setting of this PR, and I plan to resubmit a new one.

Connor1996:
> There seems to be some problem with the sign-off setting of this PR, and I plan to resubmit a new one.

No bother, it's okay now

src/table_builder.cc
Borelset: If so, there will be two AddToBaseTable functions in this class. How about AddSmallToBaseTable?

Borelset: Well, another AddToBaseTable is also adding small values.
I will rename it with AddToBaseTable.

Connor1996: emmm... let's call it `AddSmall`

Borelset:
I have reformat code and rename the function. 
The test results seems ok.

Connor1996 (approved):

src/table_builder.cc
```diff
@@ -115,16 +106,27 @@ void TitanTableBuilder::Add(const Slice& key, const Slice& value) {
...
+    AddSmall(key, ikey, value);
   } else {
-    assert(builder_unbuffered());
+    // Mainly processing kTypeMerge and kTypeBlobIndex in both flushing and
+    // compaction.
+    AddSmall(key, ikey, value);
+  }
+}
+
+void TitanTableBuilder::AddSmall(const Slice& key,
```
tabokie, line 117: The naming is still confusing. How about renaming the old `AddToBaseTable` to `AddBlobResultsToBase`, and `AddSmall` to `AddBase` (symmetric to `AddBlob`).

tabokie (commented):
Rest looks good.

src/table_builder.cc
```diff
@@ -180,7 +182,7 @@ void TitanTableBuilder::AddBlob(const ParsedInternalKey& ikey,
       cf_options_.blob_file_target_size) {
     // if blob file hit the size limit, we have to finish it
     // in this case, when calling `BlobFileBuilder::Finish`, builder will be in
-    // unbuffered state, so it will not trigger another `AddToBaseTable` call
+    // unbuffered state, so it will not trigger another `AddSmall` call
```
Connor1996, line 185: is it replaced by mistake?

Borelset: Yes, I will fix it.

src/table_builder.cc
Borelset: ok, I will do these changes.

tabokie (approved):

ti-srebot (approved):

Issue #236, opened 2022-04-21
In BlobGCJob::DoRunGC(), the `file_size` is always = 0.


apple-ouyang:
In BlobGCJob::DoRunGC(), the `file_size` is always = 0.
I guess this variable is used to accumulate the file size of blob file.
Once the `file_size >= blob_gc_->titan_cf_options().blob_file_target_size`, Titan will create a new blob file right?
But there is no file_size += blob file size code.
So should we add some code or just delte the file_size?

Connor1996:
Nice catch. Yes, the code is missing. Would you like to fix it?

apple-ouyang:
yeah, I will try to make a PR

apple-ouyang:
I have make a PR here #237

PR #237, opened 2022-04-21
fix file_size always = 0 bug

apple-ouyang:
#236

### DIFF
--- a/src/blob_gc_job.cc
+++ b/src/blob_gc_job.cc
@@ -164,8 +164,6 @@ Status BlobGCJob::DoRunGC() {
   //  uint64_t total_entry_num = 0;
   //  uint64_t total_entry_size = 0;
 
-  uint64_t file_size = 0;
-
   std::string last_key;
   bool last_key_valid = false;
   gc_iter->SeekToFirst();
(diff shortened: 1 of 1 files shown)


Connor1996 (commented):
Please add a test to cover it

apple-ouyang:
> Please add a test to cover it

I am a noob sorry. What a good test should be like? Is there a example or some docs? Or do you mean I should run the titan_*_test?

And how should I cover it? Is there a cover ratio or something that I can tell that I cover it?

Connor1996:
@apple-ouyang You can check `TEST_F(TableBuilderTest, TargetSize)`, add a test for blob gc that it would fail without your change.

apple-ouyang:
Thanks @Connor1996, I will check this code and figure out a cover test.

apple-ouyang:
I change the code. Referenced by TitanTableBuilder::AddBlob in the file table_build.cc::182.
Here is the code written in table_build.cc::182:
```c++
  if (blob_handle_->GetFile()->GetFileSize() >=
      cf_options_.blob_file_target_size) {
    // if blob file hit the size limit, we have to finish it
    // in this case, when calling `BlobFileBuilder::Finish`, builder will be in
    // unbuffered state, so it will not trigger another `AddBlobResultsToBase`
    // call
    FinishBlobFile();
  }
```
I guess my change will work and don't need a test right? Because this is how you write in the table_build.cc.
@Connor1996

Connor1996:
> I change the code. Referenced by TitanTableBuilder::AddBlob in the file table_build.cc::182. Here is the code written in table_build.cc::182:
> 
> ```c++
>   if (blob_handle_->GetFile()->GetFileSize() >=
>       cf_options_.blob_file_target_size) {
>     // if blob file hit the size limit, we have to finish it
>     // in this case, when calling `BlobFileBuilder::Finish`, builder will be in
>     // unbuffered state, so it will not trigger another `AddBlobResultsToBase`
>     // call
>     FinishBlobFile();
>   }
> ```
> 
> I guess my change will work and don't need a test right? Because this is how you write in the table_build.cc. @Connor1996

Adding a case is to make sure it won't be broken in the future. It is not about how you write the code.

Connor1996:
@apple-ouyang friendly ping

PR #240, opened 2022-05-03
Fix compute all gc score all time

apple-ouyang:


apple-ouyang:
I get a test case fail, I will fix it.

apple-ouyang:
I originally wanted to add a hash table to store files that had already been computed GCScore.  
Only new blob file calculate the gcscore. 
But there is situation when you need update the blob score.
As an example. When reopen the database, In the `Openimply::Blob_file_set_ ->Open(column_families)`, then   `Apply` is called, which in turn calls ComputeGCScore, which causes the gCscore of these files to be initialized to 0.  
That's when updates are likely to occur.  Adding a hash table to filter might also increase complexity, so I'm not going to bother, that's it. I just give up.

Issue #238, opened 2022-04-23
A status don't check and are written by later code in BlobGCJob::InstallOutputBlobFiles


apple-ouyang:
In BlobGCJob::InstallOutputBlobFiles, which is blob_gc_joob.cc:415, the status are not checked.
```C++
for (auto& builder : blob_file_builders_) {
    BlobFileBuilder::OutContexts contexts;
    s = builder.second->Finish(&contexts);  //Status not check
    BatchWriteNewIndices(contexts, &s);
    if (!s.ok()) {
      break;
    }
    metrics_.gc_num_new_files++;
```
At the same time, I think the `BatchWriteNewIndices(contexts, &s)` can be rewritten like s = BatchWriteNewIndices(contexts)
This will ensure code consistency as every funciton return a stauts.

I can make a PR to fix thoese 2 things If you like.

Connor1996:
Yes, better rewrite it.

apple-ouyang:
I push a PR here #243

PR #243, opened 2022-05-04
Check a status in InstallOutputBlobFiles and Rewrite BatchWriteNewIndices

apple-ouyang:
The issue is here #238.
I push 3 commits. 
One is add a status check.
One is change the `void BatchWriteNewIndices(BlobFileBuilder::OutContexts& contexts, Status* s)` to `Status BatchWriteNewIndices(BlobFileBuilder::OutContexts& contexts)`
One is delete MergeBlobIndex if we just don't use it when gc_merge_rewrite_ = false.

If you want, I can squash the 3 commit into 1 commit.

### DIFF
--- a/src/blob_gc_job.cc
+++ b/src/blob_gc_job.cc
@@ -246,7 +246,7 @@ Status BlobGCJob::DoRunGC() {
     BlobFileBuilder::OutContexts contexts;
     blob_file_builder->Add(blob_record, std::move(ctx), &contexts);
 
-    BatchWriteNewIndices(contexts, &s);
+    s = BatchWriteNewIndices(contexts);
 
     if (!s.ok()) {
       break;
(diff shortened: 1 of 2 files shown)


src/blob_gc_job.cc
```diff
@@ -269,46 +269,46 @@ Status BlobGCJob::DoRunGC() {
...
-
+  Status s;
+  for (const auto& ctx : contexts) {
     std::string index_entry;
     BlobIndex original_index = ctx->original_blob_index;
     ParsedInternalKey ikey;
     if (!ParseInternalKey(ctx->key, &ikey)) {
-      *s = Status::Corruption(Slice());
-      return;
+      return Status::Corruption(Slice());;
```
Connor1996, line 280: add a meaning error message

src/blob_gc_job.cc
```diff
@@ -269,46 +269,46 @@ Status BlobGCJob::DoRunGC() {
...
     }
     if (!gc_merge_rewrite_) {
-      merge_blob_index.EncodeToBase(&index_entry);
+      const BlobIndex &index = ctx->new_blob_index;
```
Connor1996, line 283: why is it different from before?

Connor1996 (commented):
sorry for the late review

src/blob_gc_job.cc
apple-ouyang: Because the merge blob index is only used in the second if branch. I just think we should only declare the merge blob index when we truly use it. This will help others understand the code easily.

src/blob_gc_job.cc
apple-ouyang: Ok, I'll add one

PR #246, opened 2022-05-12
work with 6.29.tikv

tabokie:
Changes to CI: https://github.com/PingCAP-QE/ci/pull/1112
 
Important changes:
- Titan's compaction filter must implement `IsStackedBlobDbInternalCompactionFilter`
  - Titan's filter implementation is not used for new CF. Fix: https://github.com/tikv/titan/pull/248
- Titan's new table factory breaks options validation.
  - Upstream Fix: https://github.com/facebook/rocksdb/pull/10094
- Many global variables are now made extern (https://github.com/facebook/rocksdb/commit/f1837595a395d2bcf5ff2a8d332a4793eaeb7d24)
  - Previously we mistakenly link the object file twice (when WITH_TESTS=ON), it will cause double free.
- Trivial move is disallowed in force bottommost compaction (https://github.com/facebook/rocksdb/commit/8891e9a0eb302c980792619fee641f31ecf19c86)
- RocksDB now relies on file_number to generate unique cache key (https://github.com/facebook/rocksdb/commit/0050a73a4fb1ba52d78655b271a0f09ba1fbf7fc)

Other notes:
- Titan's GC can be moved to `CompactionFilter::PrepareBlobOutput` in the future


cmake/rocksdb_flags.cmake
```diff
@@ -6,42 +6,52 @@ option(WITH_JEMALLOC "build with JeMalloc" OFF)
 if (WITH_JEMALLOC)
   find_package(JeMalloc REQUIRED)
   add_definitions(-DROCKSDB_JEMALLOC -DJEMALLOC_NO_DEMANGLE)
-  include_directories(${JEMALLOC_INCLUDE_DIR})
+  include_directories(${JeMalloc_INCLUDE_DIRS})
```
Connor1996, line 9: why have these changes

src/db_iter.cc
```diff
@@ -0,0 +1,161 @@
+#include "db_iter.h"
+
+#include "blob_storage.h"
+#include "db/arena_wrapped_db_iter.h"
+
+namespace rocksdb {
+namespace titandb {
+
+TitanDBIterator::TitanDBIterator(const TitanReadOptions &options,
```
Connor1996, line 9: why move

src/util.cc
```diff
@@ -1,5 +1,6 @@
 #include "util.h"
 
+#include "util/compression.h"
```
Connor1996, line 3: how about adding rocksdb code path with a leading `rocksdb`, so we can distinguish them easily with Titan's.

tabokie: It doesn't work like that. "ROCKSDB_DIR" and "ROCKSDB_DIR/include" are added as project include dirs. Files like "util/compression.h" reside under ROCKSDB_DIR. To include them with a leading "rocksdb", you have to include "ROCKSDB_DIR/..", which is not what we wanted.

And it's actually easy to distinguish (so far): `#include "xx.h"` and `#include "titan/xx.h"` is titan's, others are all rocksdb's.

cmake/rocksdb_flags.cmake
tabokie: We use `rocksdb/cmake/modules/*` to resolve these libraries. And the names in rocksdb is changed.

src/db_iter.cc
tabokie: Don't even remember.. I'll just revert and see.

cmake/rocksdb_flags.cmake
```diff
@@ -6,42 +6,52 @@ option(WITH_JEMALLOC "build with JeMalloc" OFF)
...
+  include_directories(${Snappy_INCLUDE_DIRS})
 endif()
 
 option(WITH_BZ2 "build with bzip2" OFF)
 if (WITH_BZ2)
-  find_package(bzip2 REQUIRED)
+  find_package(BZip2 REQUIRED)
   add_definitions(-DBZIP2)
-  include_directories(${BZIP2_INCLUDE_DIR})
+  if(BZIP2_INCLUDE_DIRS)
```
Connor1996, line 23: no need to change?

tabokie: There are both DIRS and DIR. DIR is probably from older version.

CMakeLists.txt
```diff
@@ -95,15 +91,30 @@ if (WITH_TITAN_TESTS OR WITH_TITAN_TOOLS)
...
   # -latomic is needed when building titandb_stress using clang.
   # check if libatomic exist before link
-  find_library(HAS_ATOMIC NAMES atomic libatomic.so.1 atomic.so.1)
-  if (HAS_ATOMIC)
-    link_libraries(atomic)
+  # Check if -latomic is required or not
+  if (NOT MSVC)
+    set(CMAKE_REQUIRED_FLAGS "--std=c++11")
+    CHECK_CXX_SOURCE_COMPILES("
+  #include <atomic>
```
Connor1996, line 98: add indent?

tabokie: I copied it from rocksdb, just keep it consistent.

CMakeLists.txt
```diff
@@ -120,20 +132,16 @@ if (WITH_TITAN_TESTS AND (NOT CMAKE_BUILD_TYPE STREQUAL "Release"))
...
         version_test)
-  set(TEST_LIBS
-        ${COMMON_LIBS}
-        testutillib
-        gtest)
 
   foreach(test ${TESTS})
-    add_executable(titan_${test} src/${test}.cc $<TARGET_OBJECTS:testharness>)
+    add_executable(titan_${test} src/${test}.cc)
       target_link_libraries(titan_${test} ${TEST_LIBS})
```
Connor1996, line 138: should have the same indent with below?

Connor1996 (approved):
rest LGTM

tabokie:
Ping @BusyJay

cmake/rocksdb_flags.cmake
```diff
@@ -6,42 +6,52 @@ option(WITH_JEMALLOC "build with JeMalloc" OFF)
...
+  find_package(Snappy REQUIRED)
   add_definitions(-DSNAPPY)
-  include_directories(${SNAPPY_INCLUDE_DIR})
+  include_directories(${Snappy_INCLUDE_DIRS})
 endif()
 
 option(WITH_BZ2 "build with bzip2" OFF)
 if (WITH_BZ2)
-  find_package(bzip2 REQUIRED)
+  find_package(BZip2 REQUIRED)
```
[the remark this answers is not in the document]
tabokie: > We use rocksdb/cmake/modules/* to resolve these libraries. And the names in rocksdb is changed.

ti-srebot (approved):

PR #249, opened 2022-05-26
Revert "[PCP-21] Titan GC doesn’t affect online write (#121)"

tabokie:
This reverts commit 4dc4ba89f8999c2981cf916badb088f86b4db227.

`gc_merge_rewrite` feature is not GA-ed, and there's no good reason to use it after `level_merge`. To make the situation worse, in newer versions of RocksDB, integrated BlobDB is more intertwined with merging logic, make this feature hard to maintain.


BusyJay (approved):

Connor1996 (approved):
LGTM

ti-srebot (approved):

PR #250, opened 2022-05-30
*: reuse statistics

BusyJay:
The type of `statistics` is shared_ptr, setting it means it should be used
for all metrics and queried from everywhere.


src/db_impl.cc
```diff
@@ -147,7 +147,11 @@ TitanDBImpl::TitanDBImpl(const TitanDBOptions& options,
   }
   dirname_ = db_options_.dirname;
   if (db_options_.statistics != nullptr) {
-    db_options_.statistics = titandb::CreateDBStatistics();
+    // The point of `statistics` is that it can be shared by multiple instances.
+    // So we should check if it's a qualified statistics instead of overwriting it.
+    db_options_.statistics->getTickerCount(TITAN_TICKER_ENUM_MAX - 1);
```
Connor1996, line 152: don't understand, it would panic

BusyJay: No, it won't. If a user sets `statistics` and also enable titan, then they should ensure `statistics` is implemented properly, use `rocksdb::CreateDBStatistics()` for example.

Connor1996: for `rocksdb::CreateDBStatistics()`, TITAN_TICKER_ENUM_MAX exceeds it's max enum count, and cause panic. It should use `titandb::CreateDBStatistics`

BusyJay: Interesting, vscode jump to `titandb::CreateDBStatistics`, which makes me think there is only one definition. Anyway, the correct implementation should be used and checked instead of reassigning a new value.

BusyJay:
PTAL

Connor1996 (approved):
LGTM

tabokie (approved):

ti-srebot (approved):

PR #260, opened 2022-09-06
respect `blob-run-mode` during GC

### LINKED ISSUE #259: Respect `blob-run-mode` during blob GC
## Summary

Right now blob GC will rewrite old blob records regardless of the value of `blob-run-mode`. These records will be ingested back to RocksDB by Titan's table builder, however, the live data statistics of involved blob files will not be properly updated (the statistics is only recalculated in `OnCompactionCompleted`, but not `OnFlushCompleted`). Much likely, these blob files will be left out by blob GC and not cleaned up properly.

Intuitively, we can set `discardable_ratio` to 1.0 to avoid triggering GC on blob files with some active records. But as it turns out, the GC score is not always faithful:
https://github.com/tikv/titan/blob/5d4b6b2a3217cfe78eae0843a0e9d7834c841152/src/blob_storage.cc#L218

## Proposed solution

First off, blob GC should respect `blob-run-mode`. This will not only avoid hitting this edge case (which is flushing blob index when blob-run-mode isn't normal), but also reduce useless work during Titan downgrading.

After the fix, the statistics problem still needs to be addressed. Because changes of `blob-run-mode` will not apply to on-going blob GC jobs. (This is less urgent because our official instructions to downgrade Titan update the option by reloading the TiKV nodes.)

tabokie:
Close #259

Also fix the bug that blob-run-mode changes doesn't apply to blob storage.

### DIFF
--- a/src/blob_gc_job.cc
+++ b/src/blob_gc_job.cc
@@ -18,7 +18,10 @@ class BlobGCJob::GarbageCollectionWriteCallback : public WriteCallback {
  public:
   GarbageCollectionWriteCallback(ColumnFamilyHandle* cfh, std::string&& _key,
                                  BlobIndex blob_index)
-      : cfh_(cfh), key_(std::move(_key)), blob_index_(blob_index) {
+      : cfh_(cfh),
+        key_(std::move(_key)),
+        blob_index_(blob_index),
+        read_bytes_(0) {
     assert(!key_.empty());
   }
 
@@ -190,7 +193,6 @@ Status BlobGCJob::DoRunGC() {
     if (blob_gc_->titan_cf_options().blob_run_mode ==
         TitanBlobRunMode::kFallback) {
       auto* cfh = blob_gc_->column_family_handle();
-      // Store WriteBatch for rewriting new Key-Index pairs to LSM
       GarbageCollectionWriteCallback callback(cfh, gc_iter->key().ToString(),
                                               blob_index);
       rewrite_batches_.emplace_back(
@@ -502,12 +504,17 @@ Status BlobGCJob::RewriteValidKeyToLSM() {
       break;
     }
     s = db_impl->WriteWithCallback(wo, &write_batch.first, &write_batch.second);
-    if (s.ok()) {
+    auto& new_blob_index = write_batch.second.new_blob_index;
+    bool rewritten_as_blob = (new_blob_index.blob_handle.size > 0);
+    if (s.ok() && rewritten_as_blob) {
       // count written bytes for new blob index.
       metrics_.gc_bytes_written += write_batch.first.GetDataSize();
       metrics_.gc_num_keys_relocated++;
       metrics_.gc_bytes_relocated += write_batch.second.blob_record_size();
       // Key is successfully written to LSM.
+    } else if (s.ok() && !rewritten_as_blob) {
+      metrics_.gc_num_keys_overwritten++;
+      metrics_.gc_bytes_overwritten += write_batch.second.blob_record_size();
     } else if (s.IsBusy()) {
       metrics_.gc_num_keys_overwritten++;
       metrics_.gc_bytes_overwritten += write_batch.second.blob_record_size();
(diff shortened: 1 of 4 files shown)


src/blob_gc_job.cc
```diff
@@ -484,23 +506,25 @@ Status BlobGCJob::RewriteValidKeyToLSM() {
       break;
     }
     s = db_impl->WriteWithCallback(wo, &write_batch.first, &write_batch.second);
-    if (s.ok()) {
+    auto& new_blob_index = write_batch.second.new_blob_index;
+    bool rewritten_as_blob = (new_blob_index.blob_handle.size > 0);
+    if (s.ok() && rewritten_as_blob) {
```
BusyJay, line 511: Why not
```
if (s.ok()) {
    if (rewritten_as_blob) {
   } else {
   }
} else if (s.isbusy()) {
}
```

src/blob_gc_job.cc
```diff
@@ -287,8 +309,8 @@ void BlobGCJob::BatchWriteNewIndices(BlobFileBuilder::OutContexts& contexts,
     blob_index.EncodeTo(&index_entry);
     // Store WriteBatch for rewriting new Key-Index pairs to LSM
     GarbageCollectionWriteCallback callback(cfh, ikey.user_key.ToString(),
-                                            std::move(original_index));
-    callback.value = index_entry;
+                                            original_index);
```
BusyJay, line 312: Why remove `move`?

tabokie: It's unnecessary. `move` implies the original value is no longer usable, but we want to keep using it in some cases.

BusyJay: Like what case? From what I can see, the value is assigned at every loop and not used after this line.

tabokie: L212, even though during runtime the `blob_index` isn't used afterwards, but semantically it shouldn't be moved out, that's why I change the constructor parameter type. After changing the type, it becomes confusing to keep using `move` at this line.

src/blob_storage.h
```diff
@@ -32,6 +32,7 @@ class BlobStorage {
               std::shared_ptr<BlobFileCache> _file_cache, TitanStats* stats)
       : db_options_(_db_options),
         cf_options_(_cf_options),
+        blob_run_mode_(_cf_options.blob_run_mode),
```
Connor1996, line 35: Let's get cf options from somewhere, instead of setting it specifically. Maybe need to update other options in the future.

BusyJay (approved):

src/blob_storage.h
tabokie: I feel like this needs a larger refactor. MutableCFOptions only contains blob_run_mode at the moment, and table factory already uses this approach. Apart from this case, it's tedious to get a latest cf_options in Titan. A better solution would be to only care about Titan's private options without considering keeping the other parts up-to-date with RocksDB.

src/blob_gc_job.cc
```diff
@@ -484,23 +506,27 @@ Status BlobGCJob::RewriteValidKeyToLSM() {
       break;
     }
     s = db_impl->WriteWithCallback(wo, &write_batch.first, &write_batch.second);
+    auto& new_blob_index = write_batch.second.new_blob_index;
```
Connor1996, line 509: extra space

src/blob_gc_job.cc
```diff
@@ -197,8 +203,24 @@ Status BlobGCJob::DoRunGC() {
       metrics_.gc_bytes_overwritten += blob_index.blob_handle.size;
       continue;
     }
-
-    last_key_valid = true;
+    last_key_is_fresh = true;
+
+    if (blob_gc_->titan_cf_options().blob_run_mode ==
+        TitanBlobRunMode::kFallback) {
```
Connor1996, line 209: why not just pause the gc in fallback mode

src/blob_gc_job.cc
tabokie: What do you mean?

src/blob_gc_job.cc
tabokie: Files will never become obsolete unless going through GC.

src/blob_gc_job.cc
Connor1996: oh, I read outdated code

Connor1996 (approved):
LGTM

ti-srebot (approved):

PR #281, opened 2023-07-27
Avoid holding mutex when syncing manifest

Connor1996:
Holding mutex when syncing manifest may block other places for a while, which affects tail latency quite a lot.

This is based on #274, should merge that first

### DIFF
--- a/src/blob_file_set.cc
+++ b/src/blob_file_set.cc
@@ -11,12 +11,13 @@ namespace titandb {
 const size_t kMaxFileCacheSize = 1024 * 1024;
 
 BlobFileSet::BlobFileSet(const TitanDBOptions& options, TitanStats* stats,
-                         std::atomic<bool>* initialized)
+                         std::atomic<bool>* initialized, port::Mutex* mutex)
     : dirname_(options.dirname),
       env_(options.env),
       env_options_(options),
       db_options_(options),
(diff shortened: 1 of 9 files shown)


tabokie (approved):

YuJuncen (approved):

src/blob_file_set.cc
```diff
@@ -209,25 +210,100 @@ Status BlobFileSet::WriteSnapshot(log::Writer* log) {
...
+
+  // The head of queue is the writer that is responsible for writing manifest
+  // for the group. Thw write group makes sure only one thread is writing to the
+  // manifest, otherwise the manifest will be corrupted.
+  while (!writer.done && &writer != manifest_writers_.front()) {
+    writer.cv.Wait();
+  }
+  if (writer.done) {
+    // The writer is handled by the head of queue's writer, just return here
+    return writer.status;
```
tonyxuqqi, line 248: manifest_writers_ will hold a reference of writer that is deleted  after the function exists.

tonyxuqqi (approved):

src/blob_file_set.cc
Connor1996: The writers must be popped up from `manifest_writers_` by the writer ahead of queue when it reaches here.

PR #283, opened 2023-08-29
Fix possible delete blob file twice error

Connor1996:
At start, it may encounter the error that blob file has been deleted twice
```
[server.rs:1916] ["failed to create kv engine: Storage Engine Status { code: IoError, sub_code: None, sev: NoError, state: 
\"Corruption: Blob file 49782 has been deleted twice\" }"]
```

It's introduced by #281. Previously, the `LogAndApply` was covered in the mutex as a whole critical area. With the lock optimization to avoid IO within the mutex, the `LogAndApply` itself is separated into different stages. There is a chance that both `DeleteBlobFilesInRanges` and `GC` select the same blob file and both pass `EditCollector::Seal` check, so the blob file can be deleted twice.

Here is a possible operation timing:
- GC chooses blob file 1 to rewrite it
- GC finish rewriting and begin to delete the input(blob file 1)
- GC triggers `LogAndApply` and pass the check of `EditCollector::Seal`
- GC unlocks the mutex to write edit record to manifest
- DeleteBlobFilesInRange is performed because the mutex is not held by GC now, then it chooses blob file 1 to delete
- DeleteBlobFilesInRange triggers `LogAndApply` and passes the check of `EditCollector::Seal` because GC is in writing manifest, still not apply the change to blob storage that blob file 1 is not obsolete now
- DeleteBlobFilesInRange unlocks the mutex and waits for GC to finish writing manifest
- GC finishes writing the manifest and locks the mutex to apply the change, now blob file 1 is marked as obsolete
- DeleteBlobFilesInRange finishes writing manifest and apply the change, blob file 1 is marked as obsolete again
- Titan restarts and finds the blob file 1 is deleted twice

The error is caused by Titan's paranoid check at the start, it could be simply recovered by ignoring the error. Then Titan would rewrite the manifest without the duplicated delete. And the data's correctness and integrity is still held. 

This PR fixes it by:
- Disable paranoid check by default
- In `DeleteBlobFilesInRange`, check the file state if it is marked by GC already, if then just skip it to avoid deleted twice.

### DIFF
--- a/src/blob_storage.cc
+++ b/src/blob_storage.cc
@@ -42,8 +42,10 @@ Status BlobStorage::GetBlobFilesInRanges(
          it != ((end != nullptr) ? blob_ranges_.upper_bound(*end)
                                  : blob_ranges_.end());
          it++) {
-      // The file is obsolete or being GCed, so just skip.
-      if (it->second->file_state() != BlobFileMeta::FileState::kNormal) {
+      // The file is obsolete or being processed(such as GC), for safety just
+      // skip.
+      if (it->second->file_state() != BlobFileMeta::FileState::kNormal &&
+          it->second->file_state() != BlobFileMeta::FileState::kToMerge) {
         continue;
       }
 
--- a/src/blob_storage.cc
+++ b/src/blob_storage.cc
@@ -26,9 +26,9 @@ Status BlobStorage::NewPrefetcher(uint64_t file_number,
                                     result);
 }
 
-Status BlobStorage::GetBlobFilesInRanges(const RangePtr* ranges, size_t n,
-                                         bool include_end,
-                                         std::vector<uint64_t>* files) {
+Status BlobStorage::GetBlobFilesInRanges(
+    const RangePtr* ranges, size_t n, bool include_end,
+    std::vector<std::shared_ptr<BlobFileMeta>>* files) {
   MutexLock l(&mutex_);
   for (size_t i = 0; i < n; i++) {
     const Slice* begin = ranges[i].start;
@@ -42,8 +42,11 @@ Status BlobStorage::GetBlobFilesInRanges(const RangePtr* ranges, size_t n,
          it != ((end != nullptr) ? blob_ranges_.upper_bound(*end)
                                  : blob_ranges_.end());
          it++) {
-      // Obsolete files are to be deleted, so just skip.
-      if (it->second->is_obsolete()) continue;
+      // The file is obsolete or being GCed, so just skip.
+      if (it->second->file_state() != BlobFileMeta::FileState::kNormal) {
+        continue;
+      }
+
       // The smallest and largest key of blob file meta of the old version are
       // empty, so skip.
       if (it->second->largest_key().empty() && end) continue;
@@ -51,7 +54,7 @@ Status BlobStorage::GetBlobFilesInRanges(const RangePtr* ranges, size_t n,
       if ((end == nullptr) ||
           (include_end && cmp->Compare(it->second->largest_key(), *end) <= 0) ||
           (!include_end && cmp->Compare(it->second->largest_key(), *end) < 0)) {
-        files->push_back(it->second->file_number());
+        files->push_back(it->second);
         if (!tmp.empty()) {
           tmp.append(" ");
         }
--- a/src/db_impl.cc
+++ b/src/db_impl.cc
@@ -65,6 +65,8 @@ class TitanDBImpl::FileManager : public BlobFileManager {
                                   std::unique_ptr<BlobFileHandle>>>& files)
       override {
     Status s;
+    if (files.empty()) return s;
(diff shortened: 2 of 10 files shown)


src/edit_collector.h
```diff
@@ -121,20 +128,33 @@ class EditCollector {
...
    public:
+    CFEditCollector(Logger* logger, bool _paranoid_check)
+        : paranoid_check_(_paranoid_check), info_log_(logger) {}
+
     Status AddFile(const std::shared_ptr<BlobFileMeta>& file) {
       auto number = file->file_number();
       if (added_files_.count(number) > 0) {
-        return Status::Corruption("Blob file " + ToString(number) +
-                                  " has been added twice");
+        TITAN_LOG_ERROR(info_log_,
```
tonyxuqqi, line 137: TITAN_LOG_ERROR can handle info_log_  is null right?

src/edit_collector.h
```diff
@@ -19,21 +19,28 @@ namespace titandb {
...
       return Status::Incomplete(
           "Should be not called after Sealed() is called");
 
     auto cf_id = edit.column_family_id_;
-    auto& collector = column_families_[cf_id];
+    if (column_families_.find(cf_id) == column_families_.end()) {
+      column_families_.emplace(cf_id,
+                               CFEditCollector(info_log_, paranoid_check_));
+    }
+    auto collector = column_families_.find(cf_id);
```
tonyxuqqi, line 36: It seems no need to change this line. The original one is clear.

src/edit_collector.h
```diff
@@ -121,20 +128,33 @@ class EditCollector {
...
     Status AddFile(const std::shared_ptr<BlobFileMeta>& file) {
       auto number = file->file_number();
       if (added_files_.count(number) > 0) {
-        return Status::Corruption("Blob file " + ToString(number) +
-                                  " has been added twice");
+        TITAN_LOG_ERROR(info_log_,
+                        "blob file %" PRIu64 " has been deleted twice\n",
+                        number);
+        if (paranoid_check_) {
+          return Status::Corruption("Blob file " + ToString(number) +
```
tonyxuqqi, line 141: If paranoid_check_ is false, I think it can return Status::OK immediately, as emplace will not overwrite existing key.

src/db_impl.cc
```diff
@@ -65,6 +65,8 @@ class TitanDBImpl::FileManager : public BlobFileManager {
                                   std::unique_ptr<BlobFileHandle>>>& files)
       override {
     Status s;
+    if (files.empty()) return s;
```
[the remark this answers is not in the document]
Connor1996: Otherwise the failpoint test is hard to reproduce

src/blob_storage.cc
```diff
@@ -42,16 +42,19 @@ Status BlobStorage::GetBlobFilesInRanges(const RangePtr* ranges, size_t n,
          it != ((end != nullptr) ? blob_ranges_.upper_bound(*end)
                                  : blob_ranges_.end());
          it++) {
-      // Obsolete files are to be deleted, so just skip.
-      if (it->second->is_obsolete()) continue;
+      // The file is obsolete or being GCed, so just skip.
+      if (it->second->file_state() != BlobFileMeta::FileState::kNormal) {
```
tonyxuqqi, line 46: is it better to explicitly (it->second->is_obsolete() ||  specify  == kBeingGC)  instead of a larger scope != BlobFileMeta::FileState::kNormal

src/db_impl.cc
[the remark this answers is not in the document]
tonyxuqqi: Not sure if the following code can be skipped safely. 
> s = db_->directory_->Fsync();
>     if (!s.ok()) {
>       return s;
>     }
>     {
>       MutexLock l(&db_->mutex_);
>       s = db_->blob_file_set_->LogAndApply(edit);
>       if (!s.ok()) {
>         db_->SetBGError(s);
>       }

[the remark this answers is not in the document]
Connor1996: Yes, there is no output file to be added, so no need to apply a version edit

src/edit_collector.h
Connor1996: yes

src/blob_storage.cc
Connor1996: The larger scope makes it safer, hard to reason about whether other states would meet other corner cases

tabokie (approved):

tonyxuqqi (approved):

nkg-:
Is this pulled into 7.5. If so, can you please link the PR for posterity. Here is the 6.5 PR - https://github.com/tikv/tikv/pull/15470

Connor1996:
@nkg- Yes, it's included in 7.5, see https://github.com/tikv/tikv/pull/15915

Issue #285, opened 2023-09-08
bug: key in compaction filter is internal key instead of userkey after apply IsStackedBlobDbInternalCompactionFilter


guoxiangCN:
since work with rocksdb_v6.29.5, (aka https://github.com/tikv/titan/pull/246), IsStackedBlobDbInternalCompactionFilter is implemented, which cause that the key passed to compaction filter from CompactionIterator::InvokeFilterIfNeeded is a internal key when value_type is kBlobIndex。 

place-1:
```
Slice& filter_key =
      (ikey_.type == kTypeValue ||
       !compaction_filter_->IsStackedBlobDbInternalCompactionFilter())
          ? ikey_.user_key
          : key_;
```

In TiKV, when we use compaction filter for GC, the obtained keys always have an extra portion:
Expected format:
```
MCE (userKey) + TSO
```
Actual format:
```
MCE (userKey) + TSO + (SeqNbr<<8 | ValueType::kBlobIndex)
```

tabokie:
Good catch! cc @Connor1996 @tonyxuqqi

Connor1996:
@guoxiangCN Would you like to file a PR to fix it? We can merge it now.

guoxiangCN:
The solution I can currently think of is to call ParseInternalKey in Titan's compactionfilter and pass the userkey to the user's compactionfilter.
It may also be possible to handle it in RockSDB, but I am a bit troubled by the boundary between stackable blobdb, new blobdb, and titan, so I would like to take a look at the official developer's ideas

Connor1996:
I think just let `IsStackedBlobDbInternalCompactionFilter` returns false for Titan

guoxiangCN:
okay, it me do a test

guoxiangCN:
@Connor1996 I have tried what you said, some unittest failed.
if IsStackedBlobDbInternalCompactionFilter returns false, the program will execute the logic of rocksdb integrated blobdb, which reads from the blob file, this will fail with a status of Status:: Corruption ("Unexpected blob index outside of action")

IsStackedBlobDbInternalCompactionFilter返回false时,无法区分程序的上下文是IntegratedBlobDB or Titan.

Connor1996:
I see. I thought integrated blobdb doesn't have specific logic for that. Then let's use your original idea `call ParseInternalKey in Titan's compactionfilter and pass the userkey to the user's compactionfilter.`

PR #287, opened 2023-09-13
Avoid pass internalkey to user-defined compaction filter

guoxiangCN:
Fix https://github.com/tikv/titan/issues/285


src/compaction_filter.h
```diff
@@ -37,20 +37,38 @@ class TitanCompactionFilter final : public CompactionFilter {
 
   Decision FilterV3(int level, const Slice &key, SequenceNumber seqno,
                     ValueType value_type, const Slice &value,
-                    std::string *new_value,
-                    std::string *skip_until) const override {
+                    std::string *new_value, std::string *skip_until) const
+      override {
+    Status s;
+    Slice user_key = key;
+    if (value_type == kBlobIndex) {
```
Connor1996, line 44: add a comment here

Connor1996 (approved):
rest LGTM

tabokie (approved):

src/compaction_filter.h
guoxiangCN: done

Connor1996 (commented):
Please add a test case to cover it

guoxiangCN:
done.  the code was formatted by clang-format-diff, but the jenkens-ci always failed in format check task....

guoxiangCN:
Hi, [Connor1996](https://github.com/Connor1996)  could u help merge this pr before this weekend, or else we will miss the release window.

Connor1996 (approved):
LGTM

Connor1996:
Please fix the format @guoxiangCN

guoxiangCN:
> Please fix the format @guoxiangCN

i had format with
```
clang-format -style=google -i ./src/compaction_filter.h
clang-format -style=google -i ./src/compaction_filter_test.cc
```
but the ci always failed. 
Is the logic of clang-format and clang-format-diff different?
May I ask for the correct formatting method?

Connor1996:
@guoxiangCN still has format issue

guoxiangCN:
@Connor1996 done, thks, It is indeed due to different versions of the clang format.

guoxiangCN:
@Connor1996 maybe this pr should be cherry pick into all, tikv-6.x, or else the gc logic maybe fatal

Connor1996:
@guoxiangCN TiKV's compaction filter is only added on write cf while titan is enabled on default cf. So TiKV isn't affected by this. If you have some needs, you can pick it.

Issue #291, opened 2023-11-29
Lazy load value in iterator


Connor1996:
User may call `Seek()` and `Next()` multiple times but doesn't need to get the value. So we can just load the value when `Value()` is called which saves a lot of unnecessary blob fetches.

tonyxuqqi:
I think it's a very good enhancement. @v01dstar  Could you please take this enhancement?

PR #297, opened 2023-12-22
Lazy load blob value in iterator

Connor1996:
Close https://github.com/tikv/titan/issues/291

It could greatly improve the scan performance.

### DIFF
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -249,8 +249,10 @@ struct TitanReadOptions : public ReadOptions {
   // Default: false
   bool key_only{false};
 
-  // If true, it will not abort when getting value from blob files failed and
+  // If false, it will not abort when failed to get value from blob files and
   // return an empty value.
+  //
+  // Default: true
   bool abort_on_failure{true};
 
   TitanReadOptions() = default;
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -249,6 +249,10 @@ struct TitanReadOptions : public ReadOptions {
   // Default: false
   bool key_only{false};
 
+  // If true, it will not abort when getting value from blob files failed and
+  // return an empty value.
+  bool abort_on_failure{true};
+
   TitanReadOptions() = default;
   explicit TitanReadOptions(const ReadOptions& options)
       : ReadOptions(options) {}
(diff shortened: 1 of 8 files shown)


include/titan/options.h
```diff
@@ -249,6 +249,10 @@ struct TitanReadOptions : public ReadOptions {
   // Default: false
   bool key_only{false};
 
+  // If true, it will not abort when getting value from blob files failed and
```
v01dstar, line 252: This is confusing, the explanation contradicts the naming.

include/titan/options.h
```diff
@@ -249,6 +249,12 @@ struct TitanReadOptions : public ReadOptions {
   // Default: false
   bool key_only{false};
 
+  // If false, it will not abort when failed to get value from blob files and
+  // return an empty value.
+  //
+  // Default: true
+  bool abort_on_failure{true};
```
tonyxuqqi, line 256: When is this flag set to `false`

src/db_iter.h
```diff
@@ -117,6 +93,28 @@ class TitanDBIterator : public Iterator {
...
+        hist_type = TITAN_PREV_MICROS;
+        break;
+      default:
+        hist_type = TITAN_SEEK_MICROS;
+        assert(false);
+    };
+    StopWatch sw(clock_, statistics(stats_), hist_type);
+    RecordTick(statistics(stats_), type_);
+    status_ = GetBlobValue();
+    if (!status_.ok()) {
```
tonyxuqqi, line 115: Prior to this change, when the value cannot be retrieved, the value() call fails because of assert(Valid() && !options_.key_only); 
After this change,  it returns an empty value when abort_on_failure is false. Why is this behavior change needed?

include/titan/options.h
Connor1996: when test

src/db_iter.h
Connor1996: Because rocksdb iterator assumes after calling `next()` and it's valid, the value must be available. So `value()` doesn't return status.  So here just abort. `abort_on_failure=false` is only for test

v01dstar (approved):
LGTM

tonyxuqqi (approved):

include/titan/options.h
tonyxuqqi: Got it. It's better to comment it that false is only for test.

glorv (approved):

Issue #294, opened 2023-12-06
Store BlobStorage pointer in ColumnFamilyHandle


Connor1996:
rocksdb embed `cfd` pointer in `ColumnFamilyHandleImpl`. One of the benefit is, with the handle, a read doesn't need to acquire db_mutex to obtain `cfd` pointer. Similarly, we can embed `BlobStorage` pointer in the handle to save mutex lock to obtain `BlobStorage` pointer. We can define `TitanColumnFamilyHandle` as following and return it to caller. The struct is safe to be used as `rocksdb::ColumnFamilyHandle` when calling rocksdb methods.

```
struct TitanColumnFamilyHandle : public rocksdb::ColumnFamilyHandleImpl {
    std::shared_ptr<BlobStorage> blob_storage;
};
```

Part of https://github.com/tikv/titan/issues/139

PR #298, opened 2023-12-28
Store blob storage ptr in column family handle to avoid mutex

v01dstar:
Ref #294

Store a shared ptr of `BlobStorage` in `ColumnFamilyHandle`, so that, while reading, Titan does not need to lock and fetch it every time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,19 +177,19 @@ if (WITH_TITAN_TOOLS)
   add_definitions(-DGFLAGS)
 
   add_executable(titandb_stress tools/titandb_stress.cc)
-  target_include_directories(titandb_stress PRIVATE ${gflags_INCLUDE_DIR})
+  target_include_directories(titandb_stress PRIVATE ${GFLAGS_INCLUDE_DIR})
   target_link_libraries(titandb_stress ${TOOLS_LIBS})
 
   add_executable(titandb_bench tools/db_bench.cc tools/db_bench_tool.cc)
-  target_include_directories(titandb_bench PRIVATE ${gflags_INCLUDE_DIR})
+  target_include_directories(titandb_bench PRIVATE ${GFLAGS_INCLUDE_DIR})
   target_link_libraries(titandb_bench ${TOOLS_LIBS})
 
   add_executable(titan_manifest_dump tools/manifest_dump.cc)
-  target_include_directories(titan_manifest_dump PRIVATE ${gflags_INCLUDE_DIR})
+  target_include_directories(titan_manifest_dump PRIVATE ${GFLAGS_INCLUDE_DIR})
   target_link_libraries(titan_manifest_dump ${TOOLS_LIBS})
   
   add_executable(titan_blob_file_dump tools/blob_file_dump.cc)
-  target_include_directories(titan_blob_file_dump PRIVATE ${gflags_INCLUDE_DIR})
+  target_include_directories(titan_blob_file_dump PRIVATE ${GFLAGS_INCLUDE_DIR})
   target_link_libraries(titan_blob_file_dump ${TOOLS_LIBS})
 endif()
 
--- a/src/db_impl.cc
+++ b/src/db_impl.cc
@@ -326,14 +326,25 @@ Status TitanDBImpl::OpenImpl(const std::vector<TitanCFDescriptor>& descs,
                           descs[i].options.table_factory /*base_table_factory*/,
                           titan_table_factories[i]}));
     column_families[(*handles)[i]->GetID()] = descs[i].options;
-    if (!descs[i].options.disable_auto_compactions) {
-      cf_with_compaction.push_back((*handles)[i]);
-    }
   }
-  TEST_SYNC_POINT_CALLBACK("TitanDBImpl::OpenImpl:BeforeOpenBlobFileSet", this);
   s = blob_file_set_->Open(column_families);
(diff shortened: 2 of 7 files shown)


tools/db_bench_tool.cc
v01dstar, line 1: Irrelevant to #294, but need. to be checked in to get my cmake and compiler happy.

CMakeLists.txt
v01dstar, line 1: Irrelevant to #294, but need. to be checked in to get my cmake and compiler happy.

src/titan_db_test.cc
```diff
@@ -355,21 +355,8 @@ TEST_F(TitanDBTest, Open) {
         background_job_started = true;
         assert(false);
       });
-  SyncPoint::GetInstance()->SetCallBack(
```
v01dstar, line 358: This test case does not have any special meaning, I guess. The new change breaks this, because by the time "BeforeOpenBlobFileSet", `default_cf_handle_` is not ready yet. We have other means to fix it, but the most graceful way is to remove it.

src/titan_checkpoint_test.cc
```diff
@@ -139,7 +139,7 @@ class CheckpointTest : public testing::Test {
 
   void Close() {
     for (auto h : handles_) {
-      delete h;
```
v01dstar, line 142: This does not follow the API protocol. And the new change breaks these tests due to the complicity of handling `default_cf_handle_`'s lifetime.

v01dstar:
Need to merge https://github.com/tikv/rocksdb/pull/353 in advance.

src/db_impl.cc
```diff
@@ -359,6 +370,9 @@ Status TitanDBImpl::Close() {
   Status s;
   CloseImpl();
   if (db_) {
+    if (default_cf_handle_ != nullptr) {
+      delete default_cf_handle_;
```
tonyxuqqi, line 374:
```suggestion
      delete default_cf_handle_;
      default_cf_handle_ = nullptr;
```

src/db_impl.cc
```diff
@@ -470,6 +484,16 @@ Status TitanDBImpl::CreateColumnFamilies(
...
       blob_file_set_->AddColumnFamilies(column_families);
+      for (size_t i = 0; i < handles->size(); i++) {
+        auto rocks_cf_handle =
+            static_cast_with_check<rocksdb::ColumnFamilyHandleImpl>(
+                (*handles)[i]);
+        auto titan_handle = new TitanColumnFamilyHandle(
+            *rocks_cf_handle,
+            blob_file_set_->GetBlobStorage(rocks_cf_handle->GetID()).lock());
+        (*handles)[i] = titan_handle;
+        delete rocks_cf_handle;
```
tonyxuqqi, line 495: Question: is this rocks_cf_handle be used by TitanColumnFamilyHandle anymore? As it's used as the const ColumnFamilyHandleImpl& in TitanColumnFamilyHandle()

src/db_impl.cc
```diff
@@ -326,15 +326,26 @@ Status TitanDBImpl::OpenImpl(const std::vector<TitanCFDescriptor>& descs,
...
     return s;
   }
+  for (size_t i = 0; i < handles->size(); i++) {
+    auto rocks_cf_handle =
+        static_cast_with_check<rocksdb::ColumnFamilyHandleImpl>((*handles)[i]);
+    auto titan_handle = new TitanColumnFamilyHandle(
+        *rocks_cf_handle,
+        blob_file_set_->GetBlobStorage(rocks_cf_handle->GetID()).lock());
+    (*handles)[i] = titan_handle;
+    delete rocks_cf_handle;
```
tonyxuqqi, line 341: Question: is this rocks_cf_handle be used by TitanColumnFamilyHandle anymore? As it's used as the const ColumnFamilyHandleImpl& in TitanColumnFamilyHandle()

tonyxuqqi: OK. Saw the other rocksdb PR. It's Ref() ed.

tonyxuqqi (approved):

src/db_impl.cc
```diff
@@ -326,15 +326,26 @@ Status TitanDBImpl::OpenImpl(const std::vector<TitanCFDescriptor>& descs,
                           descs[i].options.table_factory /*base_table_factory*/,
                           titan_table_factories[i]}));
     column_families[(*handles)[i]->GetID()] = descs[i].options;
-    if (!descs[i].options.disable_auto_compactions) {
-      cf_with_compaction.push_back((*handles)[i]);
```
Connor1996, line 330: why

src/db_impl.h
```diff
@@ -36,6 +52,11 @@ class TitanDBImpl : public TitanDB {
 
   Status Close() override;
 
+  using TitanDB::DefaultColumnFamily;
+  ColumnFamilyHandle* DefaultColumnFamily() const override {
```
Connor1996, line 56: why add it specifially

v01dstar: It had some code there to help with debugging, removed before proposing this PR. Removing this now.

src/db_impl.cc
v01dstar: Done

src/db_impl.h
Connor1996: seems you didn't push the commit

Connor1996 (approved):
LGTM

PR #301, opened 2024-01-04
Create new manifest after deleting obsolete files

v01dstar:
Close https://github.com/tikv/tikv/issues/16256

### DIFF
--- a/src/blob_file_set.cc
+++ b/src/blob_file_set.cc
@@ -127,7 +127,7 @@ Status BlobFileSet::Recover() {
     FileType file_type;
     if (!ParseFileName(f, &file_number, &file_type)) continue;
     if (alive_files.find(file_number) != alive_files.end()) continue;
-    // The newly created manifest file is not obsolete.
+    // We will delete the old manifest file and keep the new one.
     if (file_number == new_manifest_file_number && file_type == kDescriptorFile)
       continue;
     if (file_type != FileType::kBlobFile &&
--- a/src/blob_file_set.cc
+++ b/src/blob_file_set.cc
@@ -138,6 +138,9 @@ Status BlobFileSet::Recover() {
     env_->DeleteFile(dirname_ + "/" + f);
   }
   return Status::OK();
+}
+
+Status BlobFileSet::OpenManifest(uint64_t file_number) {
   Status s;
 
   auto file_name = DescriptorFileName(dirname_, file_number);
--- a/src/blob_file_set.cc
+++ b/src/blob_file_set.cc
@@ -127,7 +127,11 @@ Status BlobFileSet::Recover() {
     FileType file_type;
     if (!ParseFileName(f, &file_number, &file_type)) continue;
     if (alive_files.find(file_number) != alive_files.end()) continue;
-    // We will delete the old manifest file and keep the new one.
+    // Newly created manifest file cannot be simply added to alive files.
+    // It has to be dealt as a special case.
+    // Output files from last GC may not recorded in the manifest, it is
+    // possible that an orphaned blob file has the same file number as the new
+    // manifest, in this case, the orphaned blob file should be deleted.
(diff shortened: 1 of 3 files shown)


src/blob_file_set.cc
```diff
@@ -136,8 +131,8 @@ Status BlobFileSet::Recover() {
                    "Titan recovery delete obsolete file %s.", f.c_str());
     env_->DeleteFile(dirname_ + "/" + f);
   }
-
-  return Status::OK();
+  auto new_manifest_file_number = NewFileNumber();
+  return OpenManifest(new_manifest_file_number);
```
Connor1996, line 135: do not change the order, otherwise the old manifest would be already deleted when failed to create new manifest

v01dstar: PTAL

src/blob_file_set.cc
```diff
@@ -129,18 +127,17 @@ Status BlobFileSet::Recover() {
     FileType file_type;
     if (!ParseFileName(f, &file_number, &file_type)) continue;
     if (alive_files.find(file_number) != alive_files.end()) continue;
+    // The newly created manifest file is not obsolete.
```
Connor1996, line 130: Please describe why need this check

Connor1996 (commented):
Please cover it by test

src/blob_file_set.cc
```diff
@@ -129,18 +127,17 @@ Status BlobFileSet::Recover() {
...
       continue;
     TITAN_LOG_INFO(db_options_.info_log,
                    "Titan recovery delete obsolete file %s.", f.c_str());
     env_->DeleteFile(dirname_ + "/" + f);
   }
-
   return Status::OK();
-}
-
-Status BlobFileSet::OpenManifest(uint64_t file_number) {
```
tonyxuqqi, line 143: is it the typo?

tonyxuqqi:
I don't get the overall idea. To me it seems the logic has no difference. 
Before the change, it will only check 
`"if (alive_files.find(file_number) != alive_files.end()) continue;"`
The original code makes new_manifest_file_number part of alive_files.

But now, alive_files do not ocntain new_manifest_file_number, but the code check it and skip explicitly. 
` if (file_number == new_manifest_file_number && file_type == kDescriptorFile)`

v01dstar:
> I don't get the overall idea. To me it seems the logic has no difference. Before the change, it will only check `"if (alive_files.find(file_number) != alive_files.end()) continue;"` The original code makes new_manifest_file_number part of alive_files.
> 
> But now, alive_files do not ocntain new_manifest_file_number, but the code check it and skip explicitly. ` if (file_number == new_manifest_file_number && file_type == kDescriptorFile)`

It was a orphaned blob file that can not be GCed since it has the same file number as the manifest (they have different prefix and suffix though).

src/blob_file_set.cc
v01dstar: Yes, fixed.

src/blob_file_set.cc
v01dstar: Done

tonyxuqqi (approved):

src/blob_file_set.cc
```diff
@@ -129,14 +127,16 @@ Status BlobFileSet::Recover() {
     FileType file_type;
     if (!ParseFileName(f, &file_number, &file_type)) continue;
     if (alive_files.find(file_number) != alive_files.end()) continue;
+    // We will delete the old manifest file and keep the new one.
```
Connor1996, line 130:
```suggestion
    // Some GC output files may not recorded in the manifest for the last time, so there may be a blob file of as same file number as the new manifest
```

v01dstar: PTAL

Connor1996 (approved):
LGTM

PR #304, opened 2024-01-08
Fix truncated file name in log

Connor1996:
The file name of titan log is not intact
bae62de2-9ad9-4dea-b785-8fdfcb176d8e

This PR does:
- Fix the wrong truncation
- Add a leading `titan/` to the file name to make it easier to distinguish with rocksdb logs

### DIFF
--- a/src/titan_logging.h
+++ b/src/titan_logging.h
@@ -20,12 +20,12 @@
 #define TITAN_LOG_STRINGIFY(x) #x
 #define TITAN_LOG_TOSTRING(x) TITAN_LOG_STRINGIFY(x)
 #define TITAN_LOG_PREPEND_FILE_LINE(FMT) \
-  ("[%s:" TITAN_LOG_TOSTRING(__LINE__) "] " FMT)
+  ("[titan/%s:" TITAN_LOG_TOSTRING(__LINE__) "] " FMT)
 
 inline const char* TitanLogShorterFileName(const char* file) {
-  // 15 is the length of "titan_logging.h".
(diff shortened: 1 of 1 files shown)


src/titan_logging.h
```diff
@@ -20,12 +20,12 @@
...
 #define TITAN_LOG_PREPEND_FILE_LINE(FMT) \
-  ("[%s:" TITAN_LOG_TOSTRING(__LINE__) "] " FMT)
+  ("[titan/%s:" TITAN_LOG_TOSTRING(__LINE__) "] " FMT)
 
 inline const char* TitanLogShorterFileName(const char* file) {
-  // 15 is the length of "titan_logging.h".
+  // 16 is the length of "titan_logging.h".
   // If the name of this file changed, please change this number, too.
-  return file + (sizeof(__FILE__) > 15 ? sizeof(__FILE__) - 15 : 0);
+  return file + (sizeof(__FILE__) > 16 ? sizeof(__FILE__) - 16 : 0);
```
tonyxuqqi, line 28: But `titan_logging.h` actual length is 15-- the length of 16 includes the terminal char, which does not matter here. 
Is there any other char as prefix?

Connor1996: the terminal char matter here

tonyxuqqi: ok, sizeof includes the \0.

tonyxuqqi (approved):

glorv (approved):

PR #305, opened 2024-01-08
Fix small file can't be gced in fallback mode

Connor1996:
In fallback mode, only files with 100% discardable ratio, namely gc score of 1.0, would be dropped. But for small files, we deem its gc score as `blob_file_discardable_ratio`. If `blob_file_discardable_ratio` is not 1.0, small files would never be dropped.


tonyxuqqi (approved):

v01dstar (approved):
LGTM

glorv (approved):

PR #306, opened 2024-01-19
Make min-blob-size online configurable

v01dstar:


CMakeLists.txt
```diff
@@ -95,9 +95,13 @@ if (WITH_TITAN_TESTS OR WITH_TITAN_TOOLS)
     set(CMAKE_REQUIRED_FLAGS "--std=c++11")
     CHECK_CXX_SOURCE_COMPILES("
 #include <atomic>
-std::atomic<uint64_t> x(0);
+struct A {
```
Connor1996, line 98: why change

src/blob_storage.h
```diff
@@ -53,7 +54,7 @@ class BlobStorage {
 
   TitanCFOptions cf_options() {
     auto _cf_options = cf_options_;
-    _cf_options.blob_run_mode = blob_run_mode_.load();
+    cf_options_.UpdateMutableOptions(mutable_cf_options_);
```
Connor1996, line 57: `_cf_options` is not changed

src/db_impl.cc
```diff
@@ -1099,49 +1100,66 @@ Options TitanDBImpl::GetOptions(ColumnFamilyHandle* column_family) const {
   return options;
 }
 
+Status TitanDBImpl::ExtractTitanCfOptions(
```
Connor1996, line 1103: As it never returns non-okay status, why not use `changed` as return value

CMakeLists.txt
v01dstar: This was for conditionally linking `libatomic`,  atomic ops over structs larger than 16 bytes require the bin link with `libatomic`. Turns out its not a easy thing to do in `rust-rocksdb`. I now use `mutex` over lock free, changing config is not a frequent op and in the code it already acquire the mutex anyway.

src/db_impl.cc
```diff
@@ -1099,49 +1100,66 @@ Options TitanDBImpl::GetOptions(ColumnFamilyHandle* column_family) const {
...
+  // Make sure base db's SetOptions success before setting Titan's.
+  if (titan_options_changed) {
     uint32_t cf_id = column_family->GetID();
     {
       MutexLock l(&mutex_);
       assert(cf_info_.count(cf_id) > 0);
       TitanColumnFamilyInfo& cf_info = cf_info_[cf_id];
-      cf_info.titan_table_factory->SetBlobRunMode(blob_run_mode);
-      cf_info.mutable_cf_options.blob_run_mode = blob_run_mode;
+      cf_info.titan_table_factory->SetMutableCFOptions(mutable_cf_options);
```
Connor1996, line 1158: Is the mutex same as the mutex held by `SetMutableCFOptions`

v01dstar: Yes, it is the db wide mutex.

v01dstar: https://github.com/v01dstar/titan/blob/7085173919bd6ce906f68435c94ff8dd8ac75d91/src/db_impl.cc#L456

CMakeLists.txt
v01dstar: https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html More readings about why we need libatomic.

src/db_impl.cc
Connor1996: Then `SetMutableCFOptions` would lock an already locked mutex?

v01dstar: I use `mutext_->AssertHeld()`, right?

v01dstar: TableFactory uses the same mutex, so it is `mutex->AssertHeld()`, BlobStorage uses its own mutex, so it is `MutexLock l(&(blob_storage_)mutex_)`

Connor1996: Oh, I misread

Connor1996 (approved):
LGTM

src/db_impl.cc
tonyxuqqi: The caller below seems not checking the status but just the `changed` parameter

src/db_impl.cc
```diff
@@ -1099,49 +1100,66 @@ Options TitanDBImpl::GetOptions(ColumnFamilyHandle* column_family) const {
...
-                       column_family->GetName().c_str(),
-                       blob_run_mode_string.c_str());
-      }
-      opts.erase(p);
-    }
-  }
+  MutableTitanCFOptions mutable_cf_options =
+      cf_info_.at(column_family->GetID()).mutable_cf_options;
+  bool titan_options_changed = false;
+  ExtractTitanCfOptions(column_family, opts, mutable_cf_options,
```
tonyxuqqi, line 1143: Here the status is not checked.

tonyxuqqi (approved):
One minor issue. Others LGTM.

src/db_impl.cc
v01dstar: Done.

PR #309, opened 2024-01-31
Refine blob cache related code

Connor1996:
This PR does:
- Insert blob cache with bottom priority to share with block cache, should merge https://github.com/tikv/rocksdb/pull/354 first
- Utilize pinnable slice to avoid one memory copy for `Get`
- Use specified memory allocator for blob cache

### DIFF
--- a/src/blob_file_reader.cc
+++ b/src/blob_file_reader.cc
@@ -178,7 +178,8 @@ Status BlobFileReader::Get(const ReadOptions& /*options*/,
     auto cache_value = new OwnedSlice(std::move(blob));
     auto cache_size = cache_value->size() + sizeof(*cache_value);
     cache_->Insert(cache_key, cache_value, cache_size,
-                   &DeleteCacheValue<OwnedSlice>, &cache_handle);
+                   &DeleteCacheValue<OwnedSlice>, &cache_handle,
+                   Cache::Priority::BOTTOM);
     buffer->PinSlice(*cache_value, UnrefCacheHandle, cache_.get(),
                      cache_handle);
   } else {
--- a/src/blob_file_reader.cc
+++ b/src/blob_file_reader.cc
@@ -116,7 +116,8 @@ Status BlobFileReader::Open(const TitanCFOptions& options,
   reader->footer_ = footer;
   if (header.flags & BlobFileHeader::kHasUncompressionDictionary) {
     s = InitUncompressionDict(footer, reader->file_.get(),
-                              &reader->uncompression_dict_);
+                              &reader->uncompression_dict_,
+                              options.memory_allocator());
     if (!s.ok()) {
       return s;
     }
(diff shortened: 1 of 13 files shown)


src/db_impl.cc
```diff
@@ -725,10 +724,6 @@ Status TitanDBImpl::GetImpl(const ReadOptions& options,
                     options.snapshot->GetSequenceNumber(),
                     s.ToString().c_str());
   }
-  if (s.ok()) {
-    value->Reset();
-    value->PinSelf(record.value);
```
Connor1996, line 730: Here introduces one memory copy

src/blob_file_reader.cc
```diff
@@ -178,11 +182,13 @@ Status BlobFileReader::Get(const ReadOptions& /*options*/,
...
-                   &DeleteCacheValue<OwnedSlice>, &cache_handle);
-    buffer->PinSlice(*cache_value, UnrefCacheHandle, cache_.get(),
-                     cache_handle);
+                   &DeleteCacheValue<OwnedSlice>, &cache_handle,
+                   Cache::Priority::BOTTOM);
+    value->PinSlice(record->value, UnrefCacheHandle, cache_.get(),
+                    cache_handle);
   } else {
-    buffer->PinSlice(blob, OwnedSlice::CleanupFunc, blob.release(), nullptr);
+    value->PinSlice(record->value, OwnedSlice::CleanupFunc, blob.release(),
```
tonyxuqqi, line 190: Why is the record->value used instead of blob? It seems to me using blob is more straightforward. 
Same for the above PinSlice---*cache_value seems better.

v01dstar: This does not feel right. `BlobRecord::DecodeFrom()` assumes the buffer contains both key and value. While, I guess this PR wants to optimize the cache utilization by storing value only. In addition to changing the PR description and func comment, we also need to change the decoding logic of `BlobRecord` accordingly.

Connor1996: > So the new logic will return the value only and the old code returns the whole kv value pair?
Old code reads the kv value pair by `record` field. New logic still can read the kv value pair by `record` field. The `record` is just a slice refering to the pinnable slice underlying buffer. The value pinnable slice holds the underlying buffer, while as a slice I let it only read out value from it instead of the whole buffer.

>  Should the caller of Get() be aware of this change?
Previously, caller wouldn't use the buffer slice directly as they don't know the content. Key and value are read by `record` field. So no need to change and I just change the naming `buffer` to `value`

> We also need to change the decoding logic of BlobRecord accordingly.
No need. When calling `BlobRecord::DecodeFrom()`, the buffer doesn't contains both key and value.

Connor1996: > So the new logic will return the value only and the old code returns the whole kv value pair?
Old code reads the kv value pair by `record` field. New logic still can read the kv value pair by `record` field. The `record` is just a slice refering to the pinnable slice underlying buffer. The value pinnable slice holds the underlying buffer, while as a slice I let it only read out value from it instead of the whole buffer.

>  Should the caller of Get() be aware of this change?
Previously, caller wouldn't use the buffer slice directly as they don't know the content. Key and value are read by `record` field. So no need to change and I just change the naming `buffer` to `value`

> We also need to change the decoding logic of BlobRecord accordingly.
No need. When calling `BlobRecord::DecodeFrom()`, the buffer doesn't contain both key and value.

v01dstar (approved):
LGTM

tonyxuqqi (approved):

SpadeA-Tang (approved):

PR #310, opened 2024-02-04
Make discardable ratio and compression type online configurable

v01dstar:


src/db_impl.cc
```diff
@@ -1126,9 +1127,43 @@ Status TitanDBImpl::ExtractTitanCfOptions(
...
     uint64_t min_blob_size = ParseUint64(option_pos->second);
     mutable_cf_options.min_blob_size = min_blob_size;
+    TITAN_LOG_INFO(db_options_.info_log, "[%s] Set min_blob_size: %" PRIu64,
+                   column_family->GetName().c_str(), min_blob_size);
     new_options.erase(option_pos);
     changed = true;
   }
+
+  option_pos = new_options.find("blob_file_compression");
+  if (option_pos != new_options.end()) {
```
Connor1996, line 1137: The pattern is repeated, can we use a for loop?

v01dstar: Done

Connor1996 (approved):
LGTM

PR #311, opened 2024-02-04
Avoid extra mutex for getting blob file reader

Connor1996:
Should merge #309 first

### DIFF
--- a/src/blob_file_cache.cc
+++ b/src/blob_file_cache.cc
@@ -27,13 +27,13 @@ BlobFileCache::BlobFileCache(const TitanDBOptions& db_options,
 
 Status BlobFileCache::Get(const ReadOptions& options, uint64_t file_number,
                           uint64_t file_size, const BlobHandle& handle,
-                          BlobRecord* record, PinnableSlice* buffer) {
+                          BlobRecord* record, PinnableSlice* value) {
   Cache::Handle* cache_handle = nullptr;
   Status s = FindFile(file_number, file_size, &cache_handle);
   if (!s.ok()) return s;
 
   auto reader = reinterpret_cast<BlobFileReader*>(cache_->Value(cache_handle));
-  s = reader->Get(options, handle, record, buffer);
+  s = reader->Get(options, handle, record, value);
   cache_->Release(cache_handle);
   return s;
 }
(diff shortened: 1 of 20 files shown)


src/blob_file_cache.cc
```diff
@@ -66,11 +65,16 @@ Status BlobFileCache::FindFile(uint64_t file_number, uint64_t file_size,
...
   std::unique_ptr<RandomAccessFileReader> file;
+  uint64_t file_size;
   {
     std::unique_ptr<FSRandomAccessFile> f;
     auto file_name = BlobFileName(db_options_.dirname, file_number);
-    s = env_->GetFileSystem()->NewRandomAccessFile(
-        file_name, FileOptions(env_options_), &f, nullptr /*dbg*/);
+    auto fs = env_->GetFileSystem();
+
+    s = fs->GetFileSize(file_name, IOOptions(), &file_size, nullptr);
```
tonyxuqqi, line 74: Why do we bother to get the file size from again (which is a bit less efficient than passing in parameter) , when the current code already has the size in the context.

src/blob_storage.cc
```diff
@@ -8,22 +8,13 @@ namespace titandb {
 
 Status BlobStorage::Get(const ReadOptions& options, const BlobIndex& index,
                         BlobRecord* record, PinnableSlice* value) {
-  auto sfile = FindFile(index.file_number).lock();
-  if (!sfile)
-    return Status::Corruption("Missing blob file: " +
-                              std::to_string(index.file_number));
-  return file_cache_->Get(options, sfile->file_number(), sfile->file_size(),
-                          index.blob_handle, record, value);
+  return file_cache_->Get(options, index.file_number, index.blob_handle, record,
```
tonyxuqqi, line 11: It will reduce the cost of a lock, but will add a cost to get file size from handle, is it cheaper?

v01dstar: Agree, can you give some numbers?

Connor1996: Hard to tell, because it depends on how long the mutex is held by others. As the mutex is shared by multiple fields, it's better to use an operation not affected by the mutex.

At least, rocksdb does this way https://github.com/tikv/rocksdb/blob/6.29.tikv/db/blob/blob_file_reader.cc#L97

tonyxuqqi: OK. After reading the code again, I think the difference is that before this PR, the lock is accquired for every call. 
While for this PR, the getting file size only happens when cache miss. So overall it should be better.

tonyxuqqi (approved):

v01dstar (approved):
LGTM

glorv (approved):

Issue #140, opened 2020-01-31
Reduce locking for blob cache hit queries


yiwu-arbug:
## Description

Currently we query blob cache in `BlobFileReader`. It is done after getting file metadata (`BlobStorage::FindFile`) and getting the file reader from `BlobFileCache`, both of which require mutex lock. If we move blob cache query to before those two calls, we saves the two mutex lock and other overheads for a cache hit query. Demo can be find here: https://github.com/yiwu-arbug/titan/tree/cache titandb_bench shows 6% improvement in throughput for in-memory workload.

Benchmark:
```
# Fill 10G DB
./titandb_bench --db="/dev/shm/titan_bench" --use_existing_db=false --titan_min_blob_size=0 --value_size=1024 --num=10000000 --compression_type=none --benchmarks="fillseq"
# Readrandom with 20G blob cache
./titandb_bench --db="/dev/shm/titan_bench" --use_existing_db=true --titan_min_blob_size=0 --value_size=1024 --num=10000000 --compression_type=none --benchmarks="readrandom" --threads=40 --duration=300 --titan_blob_cache_size=20000000000
```
Result:
```
master: readrandom   :      65.210 micros/op 613374 ops/sec;  608.4 MB/s (4599999 of 4599999 found)
demo: readrandom   :      61.398 micros/op 651462 ops/sec;  646.1 MB/s (4823999 of 4823999 found)
```

what's left to be done on top of the demo:
- [ ] Do the same for Iterator path. The demo only change point-get path.
- [ ] Allocate blob cache memory using memory allocator: https://github.com/facebook/rocksdb/blob/master/include/rocksdb/cache.h#L78
- [ ] Remove `OwnedSlice` wrapper. Cache raw `char*` and use the deleter from `CacheAllocationPtr` as cache item deleter

## Score
3000

## SIG slack channel（must）:
`#sig-engine`

## Mentor
@Connor1996 

## Recommended Skills：
c++

skyitachi:
I use cache in Get or Iterator path, but found it will fail the `BlobFileCorruptionErrorHandling` test which change result pass through the cache by inject callback. 
I wonder if any real use case which change data directly in storage level or can I just work around this test case .

DorianZheng:
[...]

PR #312, opened 2024-02-05
Check blob cache before getting blob file reader 

Connor1996:
Close https://github.com/tikv/titan/issues/140

Move blob cache query to before getting blob file reader, then we save the two mutex lock and other overheads for a cache hit query.

### DIFF
--- a/src/blob_storage.cc
+++ b/src/blob_storage.cc
@@ -8,22 +8,13 @@ namespace titandb {
 
 Status BlobStorage::Get(const ReadOptions& options, const BlobIndex& index,
                         BlobRecord* record, PinnableSlice* value) {
-  auto sfile = FindFile(index.file_number).lock();
-  if (!sfile)
-    return Status::Corruption("Missing blob file: " +
-                              std::to_string(index.file_number));
-  return file_cache_->Get(options, sfile->file_number(), sfile->file_size(),
-                          index.blob_handle, record, value);
+  return file_cache_->Get(options, index.file_number, index.blob_handle, record,
+                          value);
 }
 
 Status BlobStorage::NewPrefetcher(uint64_t file_number,
                                   std::unique_ptr<BlobFilePrefetcher>* result) {
-  auto sfile = FindFile(file_number).lock();
-  if (!sfile)
-    return Status::Corruption("Missing blob wfile: " +
-                              std::to_string(file_number));
-  return file_cache_->NewPrefetcher(sfile->file_number(), sfile->file_size(),
-                                    result);
+  return file_cache_->NewPrefetcher(file_number, result);
 }
 
 Status BlobStorage::GetBlobFilesInRanges(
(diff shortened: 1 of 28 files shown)


src/blob_storage.cc
```diff
@@ -6,24 +6,71 @@
...
+  return cache_key;
+}
+
+Status BlobStorage::TryGetBlobCache(const std::string& cache_key,
+                                    BlobRecord* record, PinnableSlice* value,
+                                    bool* cache_hit) {
+  Status s;
+  Cache::Handle* cache_handle = blob_cache_->Lookup(cache_key);
+  if (cache_handle) {
+    *cache_hit = true;
```
tonyxuqqi, line 22: `*cache_hit = true;` should be set only when s is ok.

src/db_iter.h
```diff
@@ -154,7 +164,8 @@ class TitanDBIterator : public Iterator {
     }
 
     buffer_.Reset();
-    s = it->second->Get(options_, index.blob_handle, &record_, &buffer_);
+    OwnedSlice blob;
+    s = it->second->Get(options_, index.blob_handle, &record_, &blob);
```
tonyxuqqi, line 168: Ideally when blob_cache is not enabled, then we can still reuse the buffer_ instead of have a new OwnedSlice.

tonyxuqqi (approved):
some minor comments

src/blob_storage.cc
```diff
@@ -6,24 +6,71 @@
...
+  OwnedSlice blob;
+  Status s = file_cache_->Get(options, index.file_number, index.blob_handle,
+                              record, &blob);
+  if (!s.ok()) {
+    return s;
+  }
+
+  if (blob_cache_ && options.fill_cache) {
+    Cache::Handle* cache_handle = nullptr;
+    auto cache_value = new OwnedSlice(std::move(blob));
```
v01dstar, line 58: Maybe using `blob.release()`?

src/db_impl.cc
```diff
@@ -1486,5 +1486,16 @@ void TitanDBImpl::DumpStats() {
   log_buffer.FlushBufferToLog();
 }
 
+std::string TitanDBImpl::GenerateCachePrefix() {
+  std::string cache_prefix;
+  // Unlike block-based table, which have per-table cache prefix, we use
+  // the same cache prefix for all blob files for blob cache.
```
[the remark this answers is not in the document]
Connor1996: To differentiate with block cache

src/blob_storage.cc
Connor1996: it returns `char *` which is not what OwnedSlicce wants

v01dstar: I mean we can pass char* to `blob_cache_->Insert()`?

Connor1996: We can't because we need a deleter `char*` which is implemented by `OwnedSlice`. The other place does the same way

Connor1996:
PTAL again @v01dstar

v01dstar (approved):
LGTM

glorv (approved):

PR #323, opened 2024-07-11
Add block size for blob file

v01dstar:
This PR introduces "blocks" into Titan. This is part of the "punch hole GC" project.
We need to align blobs with blocks to be able to locate next available blobs (or iterating the blob files) when there are holes in the file.

### DIFF
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -161,6 +161,18 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // Default: false
   bool skip_value_in_compaction_filter{false};
 
+  // The size of each block in blob file. 0 means not using block-based mode,
+  // blob entries are stored continuously. While non-zero value means using
+  // block-based mode, blob entries are stored in blocks, and each block stores
+  // exactkt one blob entry, if the last block is not full, it will be filled
+  // with zeros. This is particularly useful when punch hole GC is enabled, as
+  // it can punch hole on block granularity and blob files are still
+  // self-explanatory after holes are created (imagine there is no alignment
+  // requirement for blob entries and Titan has to distinguish between real
+  // data's 0s and 0s created by punch holes).
+  uint64_t block_size{4096};
+  bool enable_punch_hole_gc{false};
+
   TitanCFOptions() = default;
   explicit TitanCFOptions(const ColumnFamilyOptions& options)
       : ColumnFamilyOptions(options) {}
@@ -205,6 +217,9 @@ struct ImmutableTitanCFOptions {
   bool level_merge;
 
   bool skip_value_in_compaction_filter;
+
+  uint64_t block_size;
+  bool enable_punch_hole_gc;
 };
 
 struct MutableTitanCFOptions {
--- a/include/titan/options.h
+++ b/include/titan/options.h
@@ -164,7 +164,7 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // The size of each block in blob file. 0 means not using block-based mode,
   // blob entries are stored continuously. While non-zero value means using
   // block-based mode, blob entries are stored in blocks, and each block stores
-  // exactkt one blob entry, if the last block is not full, it will be filled
+  // exactly one blob entry, if the last block is not full, it will be filled
   // with zeros. This is particularly useful when punch hole GC is enabled, as
   // it can punch hole on block granularity and blob files are still
   // self-explanatory after holes are created (imagine there is no alignment
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -7,6 +7,8 @@
 namespace rocksdb {
 namespace titandb {
 
+static const std::string kPad = std::string(4096, 0);
+
 BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
                                  const TitanCFOptions& cf_options,
                                  WritableFileWriter* file,
@@ -38,15 +40,18 @@ BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
 }
 
 void BlobFileBuilder::FillBlockWithPad() {
-  if (block_size_ == 0) {
+  if (block_size_ == 0 || file_->GetFileSize() % block_size_ == 0) {
     return;
   }
   size_t pad_size = block_size_ - (file_->GetFileSize() % block_size_);
-  if (pad_size == block_size_) {
-    return;
+  while (pad_size > kPad.size()) {
+    status_ = file_->Append(kPad);
+    if (!ok()) {
+      return;
+    }
+    pad_size -= kPad.size();
   }
-  std::string pad(pad_size, 0);
-  file_->Append(pad);
+  status_ = file_->Append(Slice(kPad.data(), pad_size));
 }
 
 void BlobFileBuilder::WriteHeader() {
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -7,8 +7,6 @@
 namespace rocksdb {
 namespace titandb {
 
-static const std::string kPad = std::string(4096, 0);
-
 BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
                                  const TitanCFOptions& cf_options,
                                  WritableFileWriter* file,
@@ -43,15 +41,17 @@ void BlobFileBuilder::FillBlockWithPad() {
   if (block_size_ == 0 || file_->GetFileSize() % block_size_ == 0) {
     return;
   }
+  static const std::string pad = std::string(4096, 0);
+
   size_t pad_size = block_size_ - (file_->GetFileSize() % block_size_);
-  while (pad_size > kPad.size()) {
-    status_ = file_->Append(kPad);
+  while (pad_size > pad.size()) {
+    status_ = file_->Append(pad);
     if (!ok()) {
       return;
     }
-    pad_size -= kPad.size();
+    pad_size -= pad.size();
   }
-  status_ = file_->Append(Slice(kPad.data(), pad_size));
+  status_ = file_->Append(Slice(pad.data(), pad_size));
 }
 
 void BlobFileBuilder::WriteHeader() {
@@ -61,6 +61,7 @@ void BlobFileBuilder::WriteHeader() {
     assert(blob_file_version_ >= BlobFileHeader::kVersion2);
     header.flags |= BlobFileHeader::kHasUncompressionDictionary;
   }
+  header.block_size = block_size_;
   std::string buffer;
   header.EncodeTo(&buffer);
   status_ = file_->Append(buffer);
--- a/src/blob_file_builder.cc
+++ b/src/blob_file_builder.cc
@@ -33,19 +33,35 @@ BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
     return;
 #endif
   }
+  block_size_ = cf_options.enable_punch_hole_gc ? cf_options.block_size : 0;
   WriteHeader();
+}
+
+void BlobFileBuilder::FillBlockWithPad() {
+  if (block_size_ == 0) {
+    return;
+  }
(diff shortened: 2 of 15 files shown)


include/titan/options.h
```diff
@@ -161,6 +161,18 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // Default: false
   bool skip_value_in_compaction_filter{false};
 
+  // The size of each block in blob file. 0 means not using block-based mode,
+  // blob entries are stored continuously. While non-zero value means using
+  // block-based mode, blob entries are stored in blocks, and each block stores
+  // exactkt one blob entry, if the last block is not full, it will be filled
```
zhangjinpeng87, line 167: s/exactkt/exactly

src/blob_file_builder.cc
```diff
@@ -33,19 +33,35 @@ BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
     return;
 #endif
   }
+  block_size_ = cf_options.enable_punch_hole_gc ? cf_options.block_size : 0;
```
zhangjinpeng87, line 36: I think we should sanitize the block_size in case of it is not divided by disk block size.

src/blob_file_builder.cc
```diff
@@ -33,19 +33,35 @@ BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
...
 
+void BlobFileBuilder::FillBlockWithPad() {
+  if (block_size_ == 0) {
+    return;
+  }
+  size_t pad_size = block_size_ - (file_->GetFileSize() % block_size_);
+  if (pad_size == block_size_) {
+    return;
+  }
+  std::string pad(pad_size, 0);
```
zhangjinpeng87, line 48: Can we use a static constant 4K length 0 array, and use ref as the input of append to eliminate the dynamic construction & destruction of string?

src/blob_file_iterator.cc
```diff
@@ -72,15 +72,31 @@ bool BlobFileIterator::Init() {
...
 
+  block_size_ = blob_file_header.block_size;
+
   assert(end_of_blob_record_ > BlobFileHeader::kMinEncodedLength);
   init_ = true;
   return true;
 }
 
+uint64_t BlobFileIterator::AdjustOffsetToNextBlockHead() {
+  if (block_size_ == 0) return;
```
zhangjinpeng87, line 83: return 0;

src/blob_format.cc
```diff
@@ -171,6 +172,26 @@ Status BlobFileMeta::DecodeFrom(Slice* src) {
   return Status::OK();
 }
 
+Status BlobFileMeta::DecodeFromV3(Slice* src) {
```
zhangjinpeng87, line 175: Too many duplicated code with V1&V2.

src/blob_format.h
```diff
@@ -327,28 +330,34 @@ struct BlobFileHeader {
...
   }
 
   uint64_t size() const {
-    return version == BlobFileHeader::kVersion1
-               ? BlobFileHeader::kMinEncodedLength
-               : BlobFileHeader::kMaxEncodedLength;
+    return version == BlobFileHeader::kVersion3
+               ? BlobFileHeader::kV3EncodedLength
+               : version == BlobFileHeader::kVersion2
+               ? ? BlobFileHeader::kV2EncodedLength
```
zhangjinpeng87, line 359: duplicated '?'. Please add ( ) to explicitly show the operator priority.

zhangjinpeng87 (approved):
rest LGTM

src/blob_file_iterator.cc
```diff
@@ -120,6 +136,10 @@ void BlobFileIterator::IterateForPrev(uint64_t offset) {
     status_ = decoder_.DecodeHeader(&header_buffer);
     if (!status_.ok()) return;
     total_length = kRecordHeaderSize + decoder_.GetRecordSize();
+    iterate_offset_ += total_length;
+    if (block_size_ != 0) {
+      total_length += AdjustOffsetToNextBlockHead();
```
hbisheng, line 141: Should we put this IF condition before line 139 (`iterate_offset_ += total_length;`)? I'm assuming we want the next `iterate_offset_ ` to be the next block head. 

Also, on line 145 we do `iterate_offset_ -= total_length` which I believe is to get the previous offset. If we increase total_length after line 139, it seems to me that we'll get the wrong offset on line 145. Am I missing something?

include/titan/options.h
```diff
@@ -161,6 +161,18 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // Default: false
   bool skip_value_in_compaction_filter{false};
 
+  // The size of each block in blob file. 0 means not using block-based mode,
```
hbisheng, line 164: Curious how we usually test changes on Titan. Do we rely on tests in TiKV?

src/version_edit.h
```diff
@@ -18,6 +18,7 @@ enum Tag {
   kDeletedBlobFile = 12,  // Deprecated, leave here for backward compatibility
   kAddedBlobFileV2 = 13,  // Comparing to kAddedBlobFile, it newly includes
                           // smallest_key and largest_key of blob file
+  kAddedBlobFileV3 = 14,
```
Connor1996, line 21: Please add comment here

src/blob_format.cc
Connor1996: Better keep the original logic of V1 and V2 unchanged

src/blob_file_iterator.cc
```diff
@@ -72,15 +72,31 @@ bool BlobFileIterator::Init() {
...
   }
 
+  block_size_ = blob_file_header.block_size;
+
   assert(end_of_blob_record_ > BlobFileHeader::kMinEncodedLength);
   init_ = true;
   return true;
 }
 
+uint64_t BlobFileIterator::AdjustOffsetToNextBlockHead() {
```
Connor1996, line 82: Need a test to cover it

include/titan/options.h
```diff
@@ -161,6 +161,18 @@ struct TitanCFOptions : public ColumnFamilyOptions {
...
+  // blob entries are stored continuously. While non-zero value means using
+  // block-based mode, blob entries are stored in blocks, and each block stores
+  // exactkt one blob entry, if the last block is not full, it will be filled
+  // with zeros. This is particularly useful when punch hole GC is enabled, as
+  // it can punch hole on block granularity and blob files are still
+  // self-explanatory after holes are created (imagine there is no alignment
+  // requirement for blob entries and Titan has to distinguish between real
+  // data's 0s and 0s created by punch holes).
+  uint64_t block_size{4096};
```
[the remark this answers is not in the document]
Connor1996: I'm concerned that 4K may waste too much space if most of the size of values are 1K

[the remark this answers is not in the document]
zhangjinpeng87: The default of min-blob-size of TiKV is 32K which is ok with 4K block_size here.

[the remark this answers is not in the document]
Connor1996: But we may set it to `min-blob-size` to 1KB

include/titan/options.h
v01dstar: Done

src/blob_file_builder.cc
v01dstar: Done

src/blob_file_iterator.cc
v01dstar: Done

src/blob_format.h
v01dstar: Done

src/blob_format.cc
v01dstar: The main reason is I put the new `block_size` in between existing fields, instead of the end of all fields. So, it is a little bit tricky to reuse the code

include/titan/options.h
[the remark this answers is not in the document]
v01dstar: default to 4096, because FS page / block size is also 4096. We can discuss this later maybe?

src/blob_file_builder.cc
```diff
@@ -33,19 +33,35 @@ BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
...
 }
 
+void BlobFileBuilder::FillBlockWithPad() {
+  if (block_size_ == 0) {
+    return;
+  }
+  size_t pad_size = block_size_ - (file_->GetFileSize() % block_size_);
+  if (pad_size == block_size_) {
+    return;
+  }
```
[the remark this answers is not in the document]
v01dstar: Done

include/titan/options.h
v01dstar: Added a test case

src/blob_format.h
```diff
@@ -327,28 +330,34 @@ struct BlobFileHeader {
...
                                      ToString(ver));
     }
     return Status::OK();
   }
 
   uint64_t size() const {
-    return version == BlobFileHeader::kVersion1
-               ? BlobFileHeader::kMinEncodedLength
-               : BlobFileHeader::kMaxEncodedLength;
+    return version == BlobFileHeader::kVersion3
```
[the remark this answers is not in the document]
v01dstar: Done

src/blob_file_iterator.cc
v01dstar: Done

src/version_edit.h
v01dstar: Done

src/blob_file_iterator.cc
v01dstar: `AdjustOffsetToNextBlockHead ()` modifies `iterate_offset_` internally.
`AdjustOffsetToNextBlockHead ()` is used in multiple places, so the adjustment of `iterate_offset_` better stays inside of the function. Here, I want to reuse the same code piece as much as possible, so it might be a little bit counter-intuitive

src/blob_file_builder.cc
```diff
@@ -33,19 +33,41 @@ BlobFileBuilder::BlobFileBuilder(const TitanDBOptions& db_options,
...
+    assert(blob_file_version_ >= BlobFileHeader::kVersion2);
     header.flags |= BlobFileHeader::kHasUncompressionDictionary;
   }
+  header.block_size = block_size_;
   std::string buffer;
   header.EncodeTo(&buffer);
   status_ = file_->Append(buffer);
+  if (block_size_ > 0) {
+    FillBlockWithPad();
+  }
```
hbisheng, line 70: nit: `FillBlockWithPad` already checks `block_size_` internally. Another way to write it is:
```suggestion
  MaybeFillBlockWithPad();
```

src/blob_file_iterator.cc
```diff
@@ -120,6 +136,10 @@ void BlobFileIterator::IterateForPrev(uint64_t offset) {
     status_ = decoder_.DecodeHeader(&header_buffer);
     if (!status_.ok()) return;
     total_length = kRecordHeaderSize + decoder_.GetRecordSize();
+    iterate_offset_ += total_length;
+    if (block_size_ != 0) {
```
hbisheng, line 140: nit: `AdjustOffsetToNextBlockHead` already checks `block_size_` so the if check seems unnecessary. 

**Advance**OffsetToNextBlockHead may be a better name because we never expect the offset to go back. Personally I would rename it to **MaybeAdvance**OffsetToNextBlockHead but I don't know if that would be too long...

hbisheng (approved):

Connor1996 (approved):
LGTM

include/titan/options.h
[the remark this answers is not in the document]
zhangjinpeng87: I have a idea of punch hole and also can set min-blob-size less than 4K and without space usage compromise.
For these blobs whose value size >= 16K or some threshold we align their start position to 4K and padding to their tails. For these value size less than this threshold, we store them like before in a compacted way. Punch hole just take effect for these large blobs which I believe can reclaim space and let each syscall of punch hole is valuable.

Issue #324, opened 2024-07-25
rate-limiter not limit the blobstorage read in compaction filter


mzygQAQ:
In compaction filter, for the entry that who's type = kTypeBlobIndex, a blob read will emit, this read not controlled by rate-limiter.

we set rocksdb.rate-bytes-per-sec="200M" and rate-mode="all-io", the compaction read reach 600M/s

mzygQAQ:
@v01dstar

mzygQAQ:
i made a mistake, the compact read not count the reads in compaction filter, so There are still issues in other stuff.
but theproblem in the title still exists

Connor1996:
Nice catch. Indeed, All cases for compaction filter have this problem.

mzygQAQ:
> i made a mistake, the compact read not count the reads in compaction filter, so There are still issues in other stuff. but theproblem in the title still exists

i made a mistak too, the compact read metrics in tikv actually count the blob read flow in titan compaction filter's read (blob_storage.get()).
because it's counted bya thread local IOSTATS_ADD(bytes_wriiten) in every Write function in WritableFileWriter  and Read function in RandomAccessFileReader/SequnceFileRead..., it doesn't need additional variable tracking and manual counting.

That can explain why even after adding rocksdb.rate-bytes-per-sec=200M/s&io type="all-io" to a tikv (which uses the apiV1TTL feature), Grafana's panel still shows its comparison read still exceeding 600M/s

mzygQAQ:
@Connor1996  Hello, do you have any suggestions for repair?  The execution path of blob_storage.get() is a critical path, which accepts both front-end reads and reads from compactFilter. The simplest way I can think of is to simply execute RequestToken in the filter, which doesn't seem very native

v01dstar:
We can propagate `rate_limiter_priority` all the way from compaction filter to `RandomAccessFileReader::Read`

v01dstar:
> We can propagate `rate_limiter_priority` all the way from compaction filter to `RandomAccessFileReader::Read`

I was looking at newer versions of RocksDB. In the current version 6.29. We can propagate `bool for_compaction` through to `RandomAccessFileReader::Read`, seems this can fix the issue.

PR #325, opened 2024-07-26
Propagate for_compaction from compaction filter

v01dstar:
Solving https://github.com/tikv/titan/issues/324.

This fix only works for 6.29 RocksDB. Once we upgrade RocksDB I will propose a different fix which, instead of propagating `for_compaction`, propagate `rate_limit_priority`.

### DIFF
--- a/src/blob_file_cache.h
+++ b/src/blob_file_cache.h
@@ -22,7 +22,12 @@ class BlobFileCache {
   // bytes. The provided buffer is used to store the record data, so
   // the buffer must be valid when the record is used.
   Status Get(const ReadOptions& options, uint64_t file_number,
-             const BlobHandle& handle, BlobRecord* record, OwnedSlice* buffer);
+             const BlobHandle& handle, BlobRecord* record, OwnedSlice* buffer) {
+    return Get(options, file_number, handle, record, buffer, false);
+  }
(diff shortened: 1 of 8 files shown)


src/blob_file_cache.h
```diff
@@ -22,7 +22,12 @@ class BlobFileCache {
   // bytes. The provided buffer is used to store the record data, so
   // the buffer must be valid when the record is used.
   Status Get(const ReadOptions& options, uint64_t file_number,
-             const BlobHandle& handle, BlobRecord* record, OwnedSlice* buffer);
+             const BlobHandle& handle, BlobRecord* record, OwnedSlice* buffer) {
+    return Get(options, file_number, handle, record, buffer, false);
+  }
+  Status Get(const ReadOptions& options, uint64_t file_number,
```
Connor1996, line 28: How about just setting a default value for `for_compaction`

Connor1996 (commented):
rest LGTM

mzygQAQ:
I want cherry-pick this

src/blob_file_cache.h
v01dstar: Done

Connor1996 (approved):
LGTM

zhangjinpeng87 (approved):
LGTM

PR #326, opened 2024-08-10
Add punch hole GC

v01dstar:


src/blob_format.h
```diff
@@ -275,6 +279,10 @@ class BlobFileMeta {
                 (file_size_ - kBlobMaxHeaderSize - kBlobFooterSize));
   }
   TitanInternalStats::StatsType GetDiscardableRatioLevel() const;
+  // This should be called with db mutex held.
```
Connor1996, line 282: Why?

include/titan/options.h
```diff
@@ -172,6 +172,7 @@ struct TitanCFOptions : public ColumnFamilyOptions {
   // data's 0s and 0s created by punch holes).
   uint64_t block_size{4096};
   bool enable_punch_hole_gc{false};
+  uint64_t punch_hole_threshold{4 * 1024 * 1024};
```
Connor1996, line 175: Can we let `punch_hole_threshold` == 0 indicate punch hole gc disabled

src/blob_format.h
```diff
@@ -291,6 +299,18 @@ class BlobFileMeta {
   std::string smallest_key_;
   std::string largest_key_;
 
+  // The effective size of current file. This is different from `file_size_`, as
+  // `file_size_` is the original size of the file, and does not consider space
+  // reclaimed by punch hole GC.
+  // We can't use file system's `st_blocks` to get the logical size, because
```
Connor1996, line 305: I think it's okay to get the size as `effective_file_size` after restart. The size doesn't have to be so precise. Indeed, it may have false positive for triggering punch hole GC, but it would be updated to the accurate number after the gc scan. 
Then, we can get rid of updating manifest for `effective_file_size`

src/blob_file_builder.cc
```diff
@@ -164,7 +164,12 @@ void BlobFileBuilder::FlushSampleRecords(OutContexts* out_ctx) {
 void BlobFileBuilder::WriteEncoderData(BlobHandle* handle) {
   handle->offset = file_->GetFileSize();
   handle->size = encoder_.GetEncodedSize();
-  live_data_size_ += handle->size;
+  if (block_size_ > 0) {
+    live_data_size_ +=
+        (handle->size + block_size_ - 1) / block_size_ * block_size_;
+  } else {
+    live_data_size_ += handle->size;
+  }
```
LykxSassinator, line 172: How about using `util::Roundup` to do the alignment.
```suggestion
  live_data_size_ += block_size_ == 0 ? handle->size : Roundup(handle->size, block_size_);
```

src/blob_file_size_collector.cc
```diff
@@ -57,11 +62,32 @@ Status BlobFileSizeCollector::AddUserKey(const Slice& /* key */,
...
+    uint64_t block_size = default_block_size_;
+    if (!file_block_sizes_.empty()) {
+      auto iter = file_block_sizes_.find(index.file_number);
+      if (iter != file_block_sizes_.end()) {
+        block_size = iter->second;
+      }
+    }
+    if (block_size > 0) {
+      // Align blob size with block size.
+      size = (size + block_size - 1) / block_size * block_size;
```
LykxSassinator, line 82: ditto.
```suggestion
      size = Roundup(size, block_size);
```

src/punch_hole_gc_job.cc
```diff
@@ -0,0 +1,170 @@
...
+    BlobIndex blob_index = iter->GetBlobIndex();
+    auto key = iter->key();
+    bool discardable = false;
+    s = WhetherToPunchHole(key, blob_index, &discardable);
+    if (!s.ok()) {
+      return s;
+    }
+
+    aligned_data_size = (blob_index.blob_handle.size + block_size - 1) /
+                        block_size * block_size;
```
LykxSassinator, line 80: ditto.
```suggestion
    aligned_data_size = Roundup(blob_index.blob_handle.size, block_size);
```

src/db_impl.cc
```diff
@@ -818,6 +835,24 @@ void TitanDBImpl::ReleaseSnapshot(const Snapshot* snapshot) {
   // We can record here whether the oldest snapshot is released.
   // If not, we can just skip the next round of purging obsolete files.
   db_->ReleaseSnapshot(snapshot);
+  {
+    MutexLock l(&mutex_);
+    if (pending_punch_hole_gc_ != nullptr && !punch_hole_gc_running_ &&
+        pending_punch_hole_gc_->snapshot()->GetSequenceNumber() ==
```
hbisheng, line 841: I know we don't expect the pending sequence number to be smaller than `GetOldestSnapshotSequence()`, but do we consider this change, to be more defensive? 
```suggestion
        pending_punch_hole_gc_->snapshot()->GetSequenceNumber() <=
```

hbisheng (commented):
It would be great to have a PR description to help future readers. The description could provide some background and highlight key aspects of implementation such as the introduction of the `PunchHoleGCJob` class and how the job is scheduled (when its snapshot becomes the oldest).

src/blob_gc_picker.cc
```diff
@@ -103,6 +115,47 @@ std::unique_ptr<BlobGC> BasicBlobGCPicker::PickBlobGC(
...
+                     "Blob file %" PRIu64 " no need punch hole gc",
+                     blob_file->file_number());
+      continue;
+    }
+    if (!stop_picking) {
+      blob_files.emplace_back(blob_file);
+      batch_size += blob_file->file_size();
+      if (batch_size >= cf_options_.max_gc_batch_size) {
+        // Stop pick file for this gc, but still check file for whether need
+        // trigger gc after this
```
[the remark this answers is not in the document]
v01dstar: Done

src/blob_format.h
v01dstar: Done

include/titan/options.h
v01dstar: Done

src/blob_file_builder.cc
v01dstar: Done

src/blob_file_iterator.cc
```diff
@@ -170,45 +169,51 @@ void BlobFileIterator::GetBlobRecord() {
...
+      readahead_end_offset_ = iterate_offset_;
+    }
 
-  if (readahead_end_offset_ < iterate_offset_) {
-    readahead_end_offset_ = iterate_offset_;
+    // If the record is a hole-punch record, we should continue to the next
+    // record by adjusting iterate_offset_, otherwise (not a hole-punch record),
+    // we should break the loop and return the record, iterate_offset_ is
+    // already adjusted inside GetBlobRecord() in this case.
+    if (live || !status().ok()) return;
```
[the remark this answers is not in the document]
v01dstar: For example, IO errors, In this case, we should not continue the loop.

As for whether setting `valid_` to false, I don't think it is necessary, since `valid()` (the function not the variable) considers `status_` already. So I didn't change the behavior (current code does not set `valid_` either when encountering io errors).

src/db_impl.cc
```diff
@@ -818,6 +818,24 @@ void TitanDBImpl::ReleaseSnapshot(const Snapshot* snapshot) {
   // We can record here whether the oldest snapshot is released.
   // If not, we can just skip the next round of purging obsolete files.
   db_->ReleaseSnapshot(snapshot);
+  {
+    MutexLock l(&mutex_);
```
v01dstar, line 822: It is not wise to acuqire a mutex here, since this is a hot path, almost all read requrests go through this (Titan creates a `ManagedSnapshot` implicitly). I will refactor this with atomic operations, but in a separate PR

PR #331, opened 2025-01-22
Fix close hang

v01dstar:
The original code may cause the program to hang forever. A GC job might get scheduled before closing, but does not get executed since [`JoinAllThreads()` discards all jobs that did not start running](https://github.com/tikv/rocksdb/blob/e25544481ba8fab12a4bc426181a4a890c95cecb/util/threadpool_imp.h#L29-L32).  

Fix it by checking `bg_gc_running_` to achieve the same goal but correctly.

I found it accidentally while running unit tests on my desktop Arch Linux, not sure why it worked fine in the CI.

### DIFF
--- a/src/db_impl.cc
+++ b/src/db_impl.cc
@@ -412,8 +412,8 @@ Status TitanDBImpl::CloseImpl() {
 
   {
     MutexLock l(&mutex_);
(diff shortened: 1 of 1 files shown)


Connor1996 (approved):
LGTM

glorv (approved):

PR #334, opened 2025-07-01
Add raw blob size to blob index

hbisheng:
This commit records the uncompressed size of blob values in `BlobIndex`. After this, TiKV can use this information to more accurately estimate region size, especially when the compression ratio is high.

### DIFF
--- a/src/blob_format.cc
+++ b/src/blob_format.cc
@@ -45,7 +45,7 @@ void BlobEncoder::EncodeRecord(const BlobRecord& record) {
 // EncodeSlice compresses the record and prepares the header. The compressed
 // record is stored in `record_`.
 void BlobEncoder::EncodeSlice(const Slice& record) {
-  raw_size_ = record.size(); // record the uncompressed size
+  raw_size_ = record.size();  // record the uncompressed size
 
   compressed_buffer_.clear();
   CompressionType compression;
@@ -134,8 +134,7 @@ void BlobIndex::EncodeTo(std::string* dst) const {
 
 Status BlobIndex::DecodeFrom(Slice* src) {
   unsigned char type;
-  if (!GetChar(src, &type) || type != kBlobRecord ||
-      !GetVarint64(src, &file_number)) {
+  if (!GetChar(src, &type) || !GetVarint64(src, &file_number)) {
     return Status::Corruption("BlobIndex");
   }
   Status s = blob_handle.DecodeFrom(src);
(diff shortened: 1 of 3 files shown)


src/blob_format.cc
```diff
@@ -99,12 +103,22 @@ Status BlobDecoder::DecodeRecord(Slice* src, BlobRecord* record,
...
+  PutVarint64(dst, raw_size);
 }
 
 Status BlobHandle::DecodeFrom(Slice* src) {
   if (!GetVarint64(src, &offset) || !GetVarint64(src, &size)) {
     return Status::Corruption("BlobHandle");
   }
+
+  // Optional: raw_size (may not exist in older versions)
+  if (!src->empty()) {
```
v01dstar, line 115: Using length to determine whether it is the old version or new is technically correct. However, I feel this is not a good solution in general. I noticed that in `BlobIndex` there is a enum `Type` which only has one entry. How about we utilize that byte, rename it to `Version`, use 1 to denote all existing, persisted blob indices, and add a 2 for "BlobIndexV2"? It is like a more future-proof implementation, although I know we are probably not gonna change it anymore. This is not a hard blocker, I will leave it for you to decide.

hbisheng: I totally agreed that leveraging `Type` is more future-proof, but I just found that the `BlobIndex::DecodeFrom` function has a tight assertion that `Type == kBlobRecord`. If we introduce a `kBlobRecordV2`, old code will fail to decode the blob indexes written by new code, which means we can't roll back. That seems to be a bigger risk to me. 

So for now, I plan to continue using length to indicate the version, but remove that `Type == kBlobRecord` assertion. The idea is that once this version has been fully deployed and replaces all old deployments (which have that assertion), we'll be a better position to safely introduce `kBlobRecordV2` in the future if necessary. Will that be a sensible approach? @v01dstar @Connor1996

Connor1996 (approved):
LGTM

glorv (approved):

hbisheng:
To my surprise, it turns out that even the current change breaks rollback compatibility. This is because the decoding logic for `BlobIndex` currently doesn't allow extra bytes (it didn't set `ignore_extra_bytes`), so older versions cannot decode `BlobIndex` if we increase its length. At this point I'm not sure if there's a way to fix the problem while still supporting rollbacks. 

`TitanDBIterator::GetBlobValue` function:


`DecodeInto` function:

